PR #37, opened 2023-09-22
renamed tid to txn and added IANA registry contents

tulshi:


arndt-s:
`txn` is already registered (https://www.iana.org/assignments/jwt/jwt.xhtml).
Do we need to register it again? I'm not sure what the guidance here is.

draft-tulshibagwale-oauth-transaction-tokens-03.md
```diff
@@ -549,7 +549,31 @@ The expiration time of a enclosing Txn-Token MUST NOT exceed the expiration time
...
+* Change Controller: IESG
+* Specification Document: Section {{txn-token-header}} of this specificaiton
+
+## JWT Registry Contents
+
+* Claim Name: `txn`
+* Claim Description: Unique transaction identifier
+* Change Controller: IESG
+* Specification Document: Section {{txn-token-claims}} of this specification
+
```
jricher, line 569: This is already registered by SET, we should not (can not) re-register it.
```suggestion
## JWT Registry Contents

```

arndt-s (approved):
Looks good to me

PR #41, opened 2023-10-03
Added additional text to constrain Txn-Token Lifetime.

PieterKas:
Constrain Txn-Token to be shorter than the Access Token Lifetime (upper bound on lifetime)

Editorial update to security considerations


draft-tulshibagwale-oauth-transaction-tokens-03.md
```diff
@@ -140,7 +140,7 @@ A service within a call chain may choose to replace the Txn-Token. This can typi
 To get a replacement Txn-Token, a service will request a new Txn-Token from the Txn-Token Service and provide the current Txn-Token and other parameters in the request. The Txn-Token service must exercise caution in what kinds of replacement requests it supports so as to not negate the entire value of Txn-Tokens.
 
 ## Txn-Token Lifetime
-Txn-Tokens are expected to be short-lived (order of minutes, e.g., 5 minutes), and as a result MAY be used only for the expected duration of an external invocation. If a long-running process such as an batch or offline task is involved, it can use a separate mechanism to perform the external invocation, but the resulting Txn-Token is still short-lived.
+Txn-Tokens are expected to be short-lived (order of minutes, e.g., 5 minutes), and as a result MAY be used only for the expected duration of an external invocation. A Txn-Token lifetime MUST NOT exceed the lifetime of the originally presented Access Token. If a long-running process such as an batch or offline task is involved, it can use a separate mechanism to perform the external invocation, but the resulting Txn-Token is still short-lived.
```
tulshi, line 143: This assumes that the original request had an access token, which may not always be the case (although will frequently be the case). Can we update the language to say "If the token or other credential presented to the Txn-Token service in order to issue a txn-token has an expiration time, then the txn-token MUST not exceed the lifetime of the originally presented token or credential"

PieterKas: Yes, that sounds good. Will make the change.

tulshi (approved):

PR #57, opened 2023-12-12
Token exchange profile update

gffletch:
Updated the text to more clearly profile the token exchange specification for use with Transaction Tokens.

### DIFF
--- a/draft-ietf-oauth-transaction-tokens.md
+++ b/draft-ietf-oauth-transaction-tokens.md
@@ -54,6 +54,11 @@ contributor:
   org: Microsoft
   email: arndts@microsoft.com
 
+- ins: B. Campbell
+  name: Brian Campbell
+  org: Ping Identity
+  email: bcampbell@pingidentity.com
+
 normative:
   RFC2119: # Keywords
   RFC8446: # TLS
@@ -323,7 +328,7 @@ The JWT body MUST have the following claims:
 The JWT body MAY have the following claims:
 
 #### Requester Context {#requester-context}
-The Txn-Token MAY contain an `req_ctx` claim, whose value is a JSON object the describes the requester context of the transaction. This MAY include the IP address information of the originating user, as well as information about the computational entity that requested the Txn-Token.
+The Txn-Token SHOULD contain an `req_ctx` claim, whose value is a JSON object the describes the requester context of the transaction. This MAY include the IP address information of the originating user, as well as information about the computational entity that requested the Txn-Token.
 
 The JSON value of the `req_ctx` claim MAY include any values the Txn-Token Service determines are interesting to downstream services that rely on the Txn-Token. The following claims are defined so that if they are included, they have the following meaning:
 
@@ -378,14 +383,15 @@ A workload requesting a Txn-Token must provide the Transaction Token Service wit
 To request a Txn-Token the workload invokes the /token endpoint with the following parameters:
 * `grant_type` REQUIRED. The value MUST be set to `urn:ietf:params:oauth:grant-type:token-exchange`
 * `audience` REQUIRED. The value MUST be set to the Trust Domain name
+* `scope` REQUIRED. A space-delimited list of case-sensitive strings where the value(s) MUST represent the specific purpose or intent of the transaction. 
 * `requested_token_type` REQUIRED. The value MUST be `urn:ietf:params:oauth:token-type:txn-token`
-* `subject_token` REQUIRED. The value MUST be a token representing the subject of the transaction. This could be an OAuth access_token received by an API GW requesting a Txn-Token or a JWT assertion constructed by a workload initiating a transaction.
+* `subject_token` REQUIRED. The value MUST be a token representing the subject of the transaction. This could be an OAuth access_token received by an API GW or a JWT assertion constructed by a workload initiating a transaction or another form of token as identified by `subject_token_type`.
 * `subject_token_type` REQUIRED. The value MUST indicate the type of the token present in the `subject_token` parameter
 
-The following additional parameters MUST be present in a Txn-Token Request:
+The following additional parameters MAY be present in a Txn-Token Request:
 
 * `request_context` OPTIONAL. This parameter contains a base64url encoded JSON object which represents the context of this transaction. The parameter SHOULD be present and how the Transaction Token Service uses this parameter is out of scope for this specification.
-* `authorization_details` OPTIONAL. This parameter contains a base64url encoded JSON object which represents additional details of the transaction that MUST remain immutable throughout the processing of the transaction.
+* `authz_details` OPTIONAL. This parameter contains a base64url encoded JSON object which represents additional details of the transaction that MUST remain immutable throughout the processing of the transaction by multiple workloads.
 
 The requesting workload MUST authenticate it's identity to the Transaction Token Service. The exact mechanism client authentication mechanism used is outside the scope of this specification. However, some common options are mutual TLS connections, OAuth 2.0 Bearer tokens using a client credentials token, or using the `actor_token` and `actor_token_type` parameters of the OAuth 2.0 Token Exchange specification. If using the `actor_token` and `actor_token_type` parameters, both parameters MUST be present in the request. The `actor_token` MUST autenticate the identity of the requesting workload.
 
@@ -397,20 +403,39 @@ Host: txn-token-service.trust-domain.example
 Content-Type: application/x-www-form-urlencoded
 
 grant_type=urn%3Aietf%3Aparams%3Aoauth%3Agrant-type%3Atoken-exchange
-&requested_token_type=urn%3Aietf%3Aparams%3Aoauth%3Atoken-type%3Atxn_token
+&requested_token_type=urn%3Aietf%3Aparams%3Aoauth%3Atoken-type%3Atxn-token
 &audience=http%3A%2F%2Ftrust-domain.example
+&scope=finance.watchlist.add
 &subject_token=eyJhbGciOiJFUzI1NiIsImtpZC...kdXjwhw
 &subject_token_type=urn%3Aietf%3Aparams%3Aoauth%3Atoken-type%3Aaccess_token
-&rctx=%7B%22param1%22%3A%22value1%22%2C%22param2%22%3A%22value2%22%2C%22ip_address%22%3A%2269.151.72.123%22%7D
+&request_context=eyAiaXBfYWRkcmVzcyI6ICIxMjcuMC4wLjEiLCAiY2xpZW50IjogIm1vYmlsZS1hcHAiLCAiY2xpZW50X3ZlcnNpb24iOiAidjExIiB9
 ~~~
 {: #figtxtokenrequest title="Example: Txn-Token Request"}
+
+
+## Txn-Token Request Processing
+When the Transaction Token Service receives a Txn-Token Request it MUST validate the requesting workload client authentication and determine if that workload is authorized to request the specified Txn-Token. The authorization policy for determining such issuance is out of scope for this specification. 
+
+Next, the Transaction Token Service MUST validate the `subject_token` and determine the values to specify as the `sub_id` of the issued Txn-Token. What `format` mechanism of Subject Identifiers for Security Event Tokens {{RFC9493}} is used for the issued Txn-Token is out of scope of this specification. For example, many deployments may chose to use the `iss_sub` format as described in section 3.2.3 of Subject Identifiers for Security Event Tokens {{RFC9493}} those others formats are perfectly valid as well.
+
+The Transaction Token Service MUST set the `iss` claim of the Txn-Token to it's own issuer URI.
+The Transaction Token Service MUST set the `iat` claim to the time of issuance of the Txn-Token.
+The Transaction Token Service MUST set the `aud` claim to the Trust Domain name of the Transaction Token Service.
+The Transaction Token Service MUST set the `exp` claim to the expiry time of the Txn-Token.
+The Transaction Token Service MUST set the `txn` claim to a unique ID specific to this transaction.
+
+The Transaction Token Service MUST take the value specified in the `scope` parameter of the request and copy it into the `purpose` claim of the issued Txn-Token. Additionaly, the `request_context` data SHOULD be added to the `req_ctx` object of the Txn-Token. In addition, the Transaction Token Service MUST put the authenticated requesting workload identifier in the `req_ctx` object as the `req_wl` claim.
+
+If an `authz_details` parameter is present in the Txn-Token Request, then the Transaction Token Service MUST specify the JSON object as the value of the `azd` claim.
+
+The Transaction Token Service MAY provide additional processing and verification that is outside the scope of this specification.
 
 
 ## Txn-Token Response {#txn-token-response}
 A successful response to a Txn-Token Request by a Transaction Token Service is called a Txn-Token Response. If the Transaction Token Service responds with an error, the error response is as described in Section 5.2 of {{RFC6749}}. The following describes required values of a Txn-Token Response:
 
 * The `token_type` value MUST be set to `txn_token`
-* The `access_token` value MUST be the Txn-Token
+* The `access_token` value MUST be the Txn-Token in base64url encoded form
 * The response MUST NOT include the values `expires_in`, `refresh_token` and `scope`
 
 {{figtxtokenresponse}} shows a non-normative example of a Txn-Token Response.
--- a/draft-ietf-oauth-transaction-tokens.md
+++ b/draft-ietf-oauth-transaction-tokens.md
@@ -365,24 +365,29 @@ The figure below {{figleaftxtokenbody}} shows a non-normative example of the JWT
 {: #figleaftxtokenbody title="Example: Txn-Token Body"}
 
 # Txn-Token Service
-A Txn-Token Service provides a OAuth 2.0 Token Exchange {{RFC8693}} endpoint that can respond to Txn-Token issuance requests. The token exchange requests it supports require extra parameters than those defined in the OAuth 2.0 Token Exchange {{RFC8693}} specification. The unique properties of the Txn-Token requests and responses are described below. The Txn-Token Service MAY optionally support other OAuth 2.0 endpoints and features, but that is not a requirement for it to be a Txn-Token Service.
+A Txn-Token Service provides a profile of the OAuth 2.0 Token Exchange {{RFC8693}} endpoint that can respond to Txn-Token issuance requests. In addition to profiling the OAuth 2.0 Token Exchange {{RFC8693}} specification, new parameters are also defined. The unique properties of the Txn-Token requests and responses are described below. The Txn-Token Service MAY optionally support other OAuth 2.0 endpoints and features, but that is not a requirement for it to be a Txn-Token Service.
 
-Each Trust Domain MUST have exactly one Txn-Token Service.
+Each Trust Domain MUST have exactly one logical Txn-Token Service.
 
 # Requesting Txn-Tokens
-A workload requests a Txn-Token from a Transaction Token Service using OAuth 2.0 Token Exchange {{RFC8693}}. The request to obtain a Txn-Token using this method is called a Txn-Token Request, and a successful response is called a Txn-Token Response. A Txn-Token Request is a Token Exchange Request, as described in Section 2.1 of {{RFC8693}} with additional parameters. A Txn-Token Response is a OAuth 2.0 token endpoint response, as described in Section 5 of {{RFC6749}}, where the `token_type` in the response has the value `txn_token`.
+A workload requests a Txn-Token from a Transaction Token Service using a profile of the OAuth 2.0 Token Exchange {{RFC8693}}. Txn-Tokens may be requested for both externally originating or internally originating requests. The profile describes how required and optional context can be provided to the Transaction Token Service in order for the Txn-Token to be issued. The request to obtain a Txn-Token using this method is called a Txn-Token Request, and a successful response is called a Txn-Token Response. The Txn-Token profile of the OAuth 2.0 Token Exchange {{RFC8693}} is described below.
 
 ## Txn-Token Request {#txn-token-request}
-A Txn-Token Request is an OAuth 2.0 Token Exchange Request, as described in Section 2.1 of {{RFC8693}}, with an additional parameter in the request. The following parameters are required in the Txn-Token Request by the OAuth 2.0 Token Exchange specification {{RFC8693}}:
+A workload requesting a Txn-Token must provide the Transaction Token Service with proof of it's identity (client authentication), the purpose of the Txn-Token and any additional context relating to the transaction being performed. Most of these elements are provided by the OAuth 2.0 Token Exchange specification and the rest are defined as new parameters. This profile also defines a new token type URN `urn:ieft:params:oauth:token-type:txn-token` which is used by the requesting workload to identify that it is requesting the Txn-Token Response to contain a Txn-Token.
 
-* The `audience` value MUST be set to the Trust Domain name
-* The `requested_token_type` value MUST be `urn:ietf:params:oauth:token-type:txn_token`
-* The `subject_token` value MUST be the external token received by the workload that authorized the call
-* The `subject_token_type` value MUST be present and indicate the type of the authorization token present in the `subject_token` parameter
+To request a Txn-Token the workload invokes the /token endpoint with the following parameters:
+* `grant_type` REQUIRED. The value MUST be set to `urn:ietf:params:oauth:grant-type:token-exchange`
+* `audience` REQUIRED. The value MUST be set to the Trust Domain name
+* `requested_token_type` REQUIRED. The value MUST be `urn:ietf:params:oauth:token-type:txn-token`
+* `subject_token` REQUIRED. The value MUST be a token representing the subject of the transaction. This could be an OAuth access_token received by an API GW requesting a Txn-Token or a JWT assertion constructed by a workload initiating a transaction.
+* `subject_token_type` REQUIRED. The value MUST indicate the type of the token present in the `subject_token` parameter
 
-The following additional parameter MUST be present in a Txn-Token Request:
+The following additional parameters MUST be present in a Txn-Token Request:
 
-* A parameter named `rctx` , whose value is a JSON object. This object contains the request context, i.e. any information the Transaction Token Service needs to understand the context of the incoming request
+* `request_context` OPTIONAL. This parameter contains a base64url encoded JSON object which represents the context of this transaction. The parameter SHOULD be present and how the Transaction Token Service uses this parameter is out of scope for this specification.
+* `authorization_details` OPTIONAL. This parameter contains a base64url encoded JSON object which represents additional details of the transaction that MUST remain immutable throughout the processing of the transaction.
+
+The requesting workload MUST authenticate it's identity to the Transaction Token Service. The exact mechanism client authentication mechanism used is outside the scope of this specification. However, some common options are mutual TLS connections, OAuth 2.0 Bearer tokens using a client credentials token, or using the `actor_token` and `actor_token_type` parameters of the OAuth 2.0 Token Exchange specification. If using the `actor_token` and `actor_token_type` parameters, both parameters MUST be present in the request. The `actor_token` MUST autenticate the identity of the requesting workload.
 
 {{figtxtokenrequest}} shows a non-normative example of a Txn-Token Request.
 
@@ -391,7 +396,8 @@ POST /txn-token-service/token_endpoint HTTP 1.1
 Host: txn-token-service.trust-domain.example
 Content-Type: application/x-www-form-urlencoded
 
-requested_token_type=urn%3Aietf%3Aparams%3Aoauth%3Atoken-type%3Atxn_token
+grant_type=urn%3Aietf%3Aparams%3Aoauth%3Agrant-type%3Atoken-exchange
+&requested_token_type=urn%3Aietf%3Aparams%3Aoauth%3Atoken-type%3Atxn_token
 &audience=http%3A%2F%2Ftrust-domain.example
 &subject_token=eyJhbGciOiJFUzI1NiIsImtpZC...kdXjwhw
 &subject_token_type=urn%3Aietf%3Aparams%3Aoauth%3Atoken-type%3Aaccess_token


draft-ietf-oauth-transaction-tokens.md
```diff
@@ -365,24 +371,30 @@ The figure below {{figleaftxtokenbody}} shows a non-normative example of the JWT
...
+A workload requesting a Txn-Token must provide the Transaction Token Service with proof of it's identity (client authentication), the purpose of the Txn-Token and any additional context relating to the transaction being performed. Most of these elements are provided by the OAuth 2.0 Token Exchange specification and the rest are defined as new parameters. This profile also defines a new token type URN `urn:ieft:params:oauth:token-type:txn-token` which is used by the requesting workload to identify that it is requesting the Txn-Token Response to contain a Txn-Token.
+
+To request a Txn-Token the workload invokes the /token endpoint with the following parameters:
+* `grant_type` REQUIRED. The value MUST be set to `urn:ietf:params:oauth:grant-type:token-exchange`
+* `audience` REQUIRED. The value MUST be set to the Trust Domain name
+* `scope` REQUIRED. A space-delimited list of case-sensitive strings where the value(s) MUST represent the specific purpose or intent of the transaction.
```
tulshi, line 387: `scope` is optional in a token exchange request. I'm supportive of making it required, but I'm trying to remember whether we have discussed this.

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -365,24 +371,30 @@ The figure below {{figleaftxtokenbody}} shows a non-normative example of the JWT
...
 
-* The `audience` value MUST be set to the Trust Domain name
-* The `requested_token_type` value MUST be `urn:ietf:params:oauth:token-type:txn_token`
-* The `subject_token` value MUST be the external token received by the workload that authorized the call
-* The `subject_token_type` value MUST be present and indicate the type of the authorization token present in the `subject_token` parameter
+The following additional parameters MAY be present in a Txn-Token Request:
 
-The following additional parameter MUST be present in a Txn-Token Request:
+* `request_context` OPTIONAL. This parameter contains a base64url encoded JSON object which represents the context of this transaction. The parameter SHOULD be present and how the Transaction Token Service uses this parameter is out of scope for this specification.
+* `authz_details` OPTIONAL. This parameter contains a base64url encoded JSON object which represents additional details of the transaction that MUST remain immutable throughout the processing of the transaction by multiple workloads.
```
tulshi, line 395: I'm not sure about this. I feel that the `azd` claim in the TraT should be minted by the TraT server, so the requesting service should only provide the context based on which the TraT can be minted. I feel the requesting service should not be able to dictate what goes into the TraT as the azd.

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -391,20 +403,40 @@ POST /txn-token-service/token_endpoint HTTP 1.1
...
 &subject_token=eyJhbGciOiJFUzI1NiIsImtpZC...kdXjwhw
 &subject_token_type=urn%3Aietf%3Aparams%3Aoauth%3Atoken-type%3Aaccess_token
-&rctx=%7B%22param1%22%3A%22value1%22%2C%22param2%22%3A%22value2%22%2C%22ip_address%22%3A%2269.151.72.123%22%7D
+&request_context=eyAiaXBfYWRkcmVzcyI6ICIxMjcuMC4wLjEiLCAiY2xpZW50IjogIm1vYmlsZS1hcHAiLCAiY2xpZW50X3ZlcnNpb24iOiAidjExIiB9
 ~~~
 {: #figtxtokenrequest title="Example: Txn-Token Request"}
 
 
+## Txn-Token Request Processing
+When the Transaction Token Service receives a Txn-Token Request it MUST validate the requesting workload client authentication and determine if that workload is authorized to request the specified Txn-Token. The authorization policy for determining such issuance is out of scope for this specification.
```
tulshi, line 418: Not sure what you mean by "the specified Txn-Token" here. Perhaps we can say "...determine if that workload is authorized to request Txn-Tokens with the requested values"?

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -391,20 +403,40 @@ POST /txn-token-service/token_endpoint HTTP 1.1
...
+
+The Transaction Token Service MUST set the `iss` claim of the Txn-Token to it's own issuer URI.
+The Transaction Token Service MUST set the `iat` claim to the time of issuance of the Txn-Token.
+The Transaction Token Service MUST set the `aud` claim to the Trust Domain name of the Transaction Token Service.
+The Transaction Token Service MUST set the `exp` claim to the expiry time of the Txn-Token.
+The Transaction Token Service MUST set the `txn` claim to a unique ID specific to this transaction.
+
+The Transaction Token Service MUST take the value specified in the `scope` parameter of the request and copy it into the `purpose` claim of the issued Txn-Token. Additionaly, the `request_context` data SHOULD be added to the `req_ctx` object of the Txn-Token. In addition, the Transaction Token Service MUST put the authenticated requesting workload identifier in the `req_ctx` object as the `req_wl` claim.
+
+If an `authz_details` parameter is present in the Txn-Token Request, then the Transaction Token Service MUST specify the JSON object as the value of the `azd` claim.
```
tulshi, line 430: I'd like to discuss this part. I feel the requesting service should not oblige the TraT service to have specific content in the TraT.

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -391,20 +403,40 @@ POST /txn-token-service/token_endpoint HTTP 1.1
...
+The Transaction Token Service MAY provide additional processing and verification that is outside the scope of this specification.
+
+
 ## Txn-Token Response {#txn-token-response}
 A successful response to a Txn-Token Request by a Transaction Token Service is called a Txn-Token Response. If the Transaction Token Service responds with an error, the error response is as described in Section 5.2 of {{RFC6749}}. The following describes required values of a Txn-Token Response:
 
 * The `token_type` value MUST be set to `txn_token`
-* The `access_token` value MUST be the Txn-Token
+* The `access_token` value MUST be the Txn-Token in base64url encoded form
 * The response MUST NOT include the values `expires_in`, `refresh_token` and `scope`
```
bc-pi, line 440: missing `issued_token_type` in this list

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -391,20 +403,40 @@ POST /txn-token-service/token_endpoint HTTP 1.1
...
+
+If an `authz_details` parameter is present in the Txn-Token Request, then the Transaction Token Service MUST specify the JSON object as the value of the `azd` claim.
+
+The Transaction Token Service MAY provide additional processing and verification that is outside the scope of this specification.
+
+
 ## Txn-Token Response {#txn-token-response}
 A successful response to a Txn-Token Request by a Transaction Token Service is called a Txn-Token Response. If the Transaction Token Service responds with an error, the error response is as described in Section 5.2 of {{RFC6749}}. The following describes required values of a Txn-Token Response:
 
 * The `token_type` value MUST be set to `txn_token`
```
bc-pi, line 438: check the way `token_type` is defined, which is kinda awakrd in this context. See https://datatracker.ietf.org/doc/html/rfc8693#section-2.2.1-2.4 and https://www.rfc-editor.org/rfc/rfc6749#section-7.1 

just using `N_A` here might be appropriate

draft-ietf-oauth-transaction-tokens.md
bc-pi, line 440: missing `issued_token_type` in this list

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -391,20 +403,40 @@ POST /txn-token-service/token_endpoint HTTP 1.1
...
-* The `access_token` value MUST be the Txn-Token
+* The `access_token` value MUST be the Txn-Token in base64url encoded form
```
bc-pi, line 439: I would not say "in base64url encoded from" here as might suggest the Txn-Token is supposed to be re-encoded or otherwise confuse folks. The Txn-Token is a JWT. Nothing more needs to be said here.

draft-ietf-oauth-transaction-tokens.md
bc-pi: also the example below is missing token_type

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -365,24 +371,30 @@ The figure below {{figleaftxtokenbody}} shows a non-normative example of the JWT
...
+
+To request a Txn-Token the workload invokes the /token endpoint with the following parameters:
+* `grant_type` REQUIRED. The value MUST be set to `urn:ietf:params:oauth:grant-type:token-exchange`
+* `audience` REQUIRED. The value MUST be set to the Trust Domain name
+* `scope` REQUIRED. A space-delimited list of case-sensitive strings where the value(s) MUST represent the specific purpose or intent of the transaction.
+* `requested_token_type` REQUIRED. The value MUST be `urn:ietf:params:oauth:token-type:txn-token`
+* `subject_token` REQUIRED. The value MUST be a token representing the subject of the transaction. This could be an OAuth access_token received by an API GW or a JWT assertion constructed by a workload initiating a transaction or another form of token as identified by `subject_token_type`.
```
bc-pi, line 389: GW probably needs to be spelled out

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -365,24 +371,30 @@ The figure below {{figleaftxtokenbody}} shows a non-normative example of the JWT
...
+* `authz_details` OPTIONAL. This parameter contains a base64url encoded JSON object which represents additional details of the transaction that MUST remain immutable throughout the processing of the transaction by multiple workloads.
 
-* A parameter named `rctx` , whose value is a JSON object. This object contains the request context, i.e. any information the Transaction Token Service needs to understand the context of the incoming request
+The requesting workload MUST authenticate it's identity to the Transaction Token Service. The exact mechanism client authentication mechanism used is outside the scope of this specification. However, some common options are mutual TLS connections, OAuth 2.0 Bearer tokens using a client credentials token, or using the `actor_token` and `actor_token_type` parameters of the OAuth 2.0 Token Exchange specification. If using the `actor_token` and `actor_token_type` parameters, both parameters MUST be present in the request. The `actor_token` MUST autenticate the identity of the requesting workload.
```
bc-pi, line 397: "OAuth 2.0 Bearer tokens using a client credentials token" -> token exchange is defined at the token endpoint, which can directly accept oauth client authentication. Also "client credentials token" isn't a defined thing (I think I know what you mean but I'm guessing).

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -365,24 +371,30 @@ The figure below {{figleaftxtokenbody}} shows a non-normative example of the JWT
...
 
 # Requesting Txn-Tokens
-A workload requests a Txn-Token from a Transaction Token Service using OAuth 2.0 Token Exchange {{RFC8693}}. The request to obtain a Txn-Token using this method is called a Txn-Token Request, and a successful response is called a Txn-Token Response. A Txn-Token Request is a Token Exchange Request, as described in Section 2.1 of {{RFC8693}} with additional parameters. A Txn-Token Response is a OAuth 2.0 token endpoint response, as described in Section 5 of {{RFC6749}}, where the `token_type` in the response has the value `txn_token`.
+A workload requests a Txn-Token from a Transaction Token Service using a profile of the OAuth 2.0 Token Exchange {{RFC8693}}. Txn-Tokens may be requested for both externally originating or internally originating requests. The profile describes how required and optional context can be provided to the Transaction Token Service in order for the Txn-Token to be issued. The request to obtain a Txn-Token using this method is called a Txn-Token Request, and a successful response is called a Txn-Token Response. The Txn-Token profile of the OAuth 2.0 Token Exchange {{RFC8693}} is described below.
```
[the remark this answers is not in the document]
gffletch: I was intending that MAY to normative. Does it make sense to be normative?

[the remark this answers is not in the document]
gffletch: I wasn't intending the text to imply that token exchange is optional, but rather that the Transaction Token spec is profiling the token exchange spec and that profile MUST be followed for the purpose of requesting Txn-Tokens. For me a profile doesn't mean the initial spec is optional but rather that the use of the transaction token spec is modified in some way.

draft-ietf-oauth-transaction-tokens.md
gffletch: I added it as required, because we have the 'purpose' claim of the Txn-Token and the requesting workload needs to specify what that purpose is. I discussed with Brian via email and we thought using the existing 'scope' parameter was a better option than defining a new parameter. I'm open to other perspectives.

draft-ietf-oauth-transaction-tokens.md
gffletch: So if all that `azd` data is in the access_token that's fine but what about the case where the requesting workload doesn't have an access token and is requesting a TraT to be minted. In that case, how does the workload ask for the additional transaction data to be added to the Trat in the `azd` claim?

draft-ietf-oauth-transaction-tokens.md
gffletch: That's clearer. I'll update with that wording.

draft-ietf-oauth-transaction-tokens.md
gffletch: I think I was wondering about that when I wrote it so happy to have a discussion about this. I'm thinking mostly about the case where an internal workload (e.g. like a scheduled job) needs a TraT and has additional details about the transaction that it wants to make immutable. We can say that it is up to the Transaction Token Service (TTS) to determine what data is included in the TraT. How the TTS makes that determination is out of scope for the specification.

draft-ietf-oauth-transaction-tokens.md
gffletch: Yes, I agree that N_A is probably better. I'll update accordingly.

draft-ietf-oauth-transaction-tokens.md
gffletch: Thanks!

draft-ietf-oauth-transaction-tokens.md
gffletch: Ok. I wasn't sure how JWT are specified as values of a JSON claim. If multiple encodings are supported that's fine.

draft-ietf-oauth-transaction-tokens.md
gffletch: Ok, let me re-work this paragraph and just point to the client authentication mechanisms supported by the /token endpoint as defined in existing RFCs.

draft-ietf-oauth-transaction-tokens.md
naveencm4u1: Since the transaction token is passed to multiple workloads in a transaction, is it ok to expose `azd` to all the workloads? Is there any guidance on which type of data can be added to `azd`.
We are working on adding common fields to `azd` and also encrypt them. Access to the decryption secret is limited to specific workloads that need the data.

gffletch: That is a really interesting question and I wonder if selective disclosure JWTs could be a way to address that. I suppose if different workloads need access to different aspects of the `azd` data that could be more complicated.

One question, we may want to consider for this specification is whether that level of detail should be left up the specific deployment and become out of scope for this specification. Regardless, it's an interesting question. Would you mind filing an issue on it? I think it needs it's own discussion.

draft-ietf-oauth-transaction-tokens.md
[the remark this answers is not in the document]
gffletch: I updated the previous paragraph (#Txn-Token Service) to be much stronger. Do we feel we need both? I can update here as well.

draft-ietf-oauth-transaction-tokens.md
gffletch: I filed an issue regarding this topic

draft-ietf-oauth-transaction-tokens.md
gffletch: I filed an issue on this topic

draft-ietf-oauth-transaction-tokens.md
gffletch: I filed an issue on this topic

draft-ietf-oauth-transaction-tokens.md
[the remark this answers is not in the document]
tulshi: not sure - it could make sense either way.

draft-ietf-oauth-transaction-tokens.md
bc-pi: Wait, what? This isn't a claim and there's only one encoding of a JWT.  This confused me :/ But "Txn-Token JWT" is in the latest commit works for me. Maybe someday we can discuss and hash out some of the confusion here. But latest text is good.

draft-ietf-oauth-transaction-tokens.md
tulshi: Issue #58

draft-ietf-oauth-transaction-tokens.md
tulshi: Issue #61

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -391,20 +403,41 @@ POST /txn-token-service/token_endpoint HTTP 1.1
...
-&rctx=%7B%22param1%22%3A%22value1%22%2C%22param2%22%3A%22value2%22%2C%22ip_address%22%3A%2269.151.72.123%22%7D
+&request_context=eyAiaXBfYWRkcmVzcyI6ICIxMjcuMC4wLjEiLCAiY2xpZW50IjogIm1vYmlsZS1hcHAiLCAiY2xpZW50X3ZlcnNpb24iOiAidjExIiB9
 ~~~
 {: #figtxtokenrequest title="Example: Txn-Token Request"}
 
 
+## Txn-Token Request Processing
+When the Transaction Token Service receives a Txn-Token Request it MUST validate the requesting workload client authentication and determine if that workload is authorized to obtain the Txn-Tokens with the requested values. The authorization policy for determining such issuance is out of scope for this specification.
+
+Next, the Transaction Token Service MUST validate the `subject_token` and determine the values to specify as the `sub_id` of the issued Txn-Token. What `format` mechanism of Subject Identifiers for Security Event Tokens {{RFC9493}} is used for the issued Txn-Token is out of scope of this specification. For example, many deployments may chose to use the `iss_sub` format as described in section 3.2.3 of Subject Identifiers for Security Event Tokens {{RFC9493}} those others formats are perfectly valid as well.
```
obfuscoder, line 420: "For example, many deployments may chose ... those others formats are perfectly valid as well."

I believe that should be two sentences or separated?

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -391,20 +403,41 @@ POST /txn-token-service/token_endpoint HTTP 1.1
...
+
+The Transaction Token Service MUST set the `iss` claim of the Txn-Token to it's own issuer URI.
+The Transaction Token Service MUST set the `iat` claim to the time of issuance of the Txn-Token.
+The Transaction Token Service MUST set the `aud` claim to the Trust Domain name of the Transaction Token Service.
+The Transaction Token Service MUST set the `exp` claim to the expiry time of the Txn-Token.
+The Transaction Token Service MUST set the `txn` claim to a unique ID specific to this transaction.
+
+The Transaction Token Service MUST take the value specified in the `scope` parameter of the request and copy it into the `purp` claim of the issued Txn-Token. Additionaly, the `request_context` data SHOULD be added to the `req_ctx` object of the Txn-Token. In addition, the Transaction Token Service MUST put the authenticated requesting workload identifier in the `req_ctx` object as the `req_wl` claim.
```
obfuscoder, line 428: The PR  suggests that the scope is simply copied to the purp claim. Is this really the case? Shouldn't rather the scope of the presented Access Token be narrowed down to what is given in the scope request param?

draft-ietf-oauth-transaction-tokens.md
gffletch: After additional discussion, we've decided to move away from using the `sub_id` construct and just use `sub` and bind the uniqueness of the `sub` value to the trust domain as defined by the `aud` claim. An updated commit will make this change.

draft-ietf-oauth-transaction-tokens.md
gffletch: Good catch. The transaction token service (TTS) is authoritative for setting the `purp` claim. The text will be updated to allow the TTS to use the `scope` value to make the determination. Note that not all transaction token requests will contain an access token.

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -391,20 +403,41 @@ POST /txn-token-service/token_endpoint HTTP 1.1
...
 
 
+## Txn-Token Request Processing
+When the Transaction Token Service receives a Txn-Token Request it MUST validate the requesting workload client authentication and determine if that workload is authorized to obtain the Txn-Tokens with the requested values. The authorization policy for determining such issuance is out of scope for this specification.
+
+Next, the Transaction Token Service MUST validate the `subject_token` and determine the values to specify as the `sub_id` of the issued Txn-Token. What `format` mechanism of Subject Identifiers for Security Event Tokens {{RFC9493}} is used for the issued Txn-Token is out of scope of this specification. For example, many deployments may chose to use the `iss_sub` format as described in section 3.2.3 of Subject Identifiers for Security Event Tokens {{RFC9493}} those others formats are perfectly valid as well.
+
+The Transaction Token Service MUST set the `iss` claim of the Txn-Token to it's own issuer URI.
+The Transaction Token Service MUST set the `iat` claim to the time of issuance of the Txn-Token.
+The Transaction Token Service MUST set the `aud` claim to the Trust Domain name of the Transaction Token Service.
```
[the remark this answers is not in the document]
gffletch: That could be a possible deployment solution provided all transaction token requests have a way to define the trust domain to be used for that request. I've updated the text to allow for this case.

draft-ietf-oauth-transaction-tokens.md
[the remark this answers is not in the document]
KramNosaj01: where to put the tokens?im sorry im just beginner

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -306,34 +312,47 @@ In the JWT Header:
...
 {: #figtxtokenheader title="Example: Txn-Token Header"}
 
-## JWT Body {#txn-token-body}
+## JWT Body Claims {#txn-token-claims}
+
+The transaction token body follows the JWT format and includes existing
+JWT claims as well as defines new claims. These claims are described below:
+
+`iss`:
+: OPTIONAL The `iss` claim as defined in {{RFC7519}} is not required as Txn-Tokens are bound to a single trust domain as defined by the `aud` claim and often the signing keys are known. The `iss` claim MAY be used in cases where the signing keys are not predetermined or it is desired that the Txn-Token Service signs with unique keys.
```
tulshi, line 321: I would like to make this language a bit more normative. How about something like "The iss claim ... is optional. In situations where there are multiple Txn-Token issuers for the same trust domain, the iss claim MUST be included and MUST describe the issuer of the Txn-Token."

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -306,34 +312,47 @@ In the JWT Header:
...
-The JWT body MAY have the following claims:
+`purp`:
+: REQUIRED A string defining the purpose or intent of this transaction.
 
-#### Requester Context {#requester-context}
-The Txn-Token MAY contain an `req_ctx` claim, whose value is a JSON object the describes the requester context of the transaction. This MAY include the IP address information of the originating user, as well as information about the computational entity that requested the Txn-Token.
+`azd`:
+: OPTIONAL A JSON object that conatains values that remain immutable throughout the call chain.
+
+`req_ctx`:
```
tulshi, line 344: For compactness, should we change this name to "rctx" (since we are not using that in the Txn-Token request anymore) or something shorter?

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -306,34 +312,47 @@ In the JWT Header:
...
 
 * `req_ip` The IP address of the requester. This MAY be the end-user or a robotic process that requested the Transaction
 * `authn` The authentication method used to idenitfy the requester. Its value is a URN that uniquely identifies the method used.
 * `req_wl` The requesting workload. A URN that uniquely identifies the computational entity that requested the Txn-Token. This entity MUST be within the Trust Domain of the Txn-Token.
 
-#### Purpose {#purpose}
-The Txn-Token MAY contain a `purp` claim, whose value specifies the purpose of the transaction. The format of this claim is a JSON string.
-
 ### Example
 The figure below {{figleaftxtokenbody}} shows a non-normative example of the JWT body of a Txn-Token:
```
tulshi, line 357: Figure needs to be updated.

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -365,24 +384,30 @@ The figure below {{figleaftxtokenbody}} shows a non-normative example of the JWT
...
 
 ## Txn-Token Request {#txn-token-request}
-A Txn-Token Request is an OAuth 2.0 Token Exchange Request, as described in Section 2.1 of {{RFC8693}}, with an additional parameter in the request. The following parameters are required in the Txn-Token Request by the OAuth 2.0 Token Exchange specification {{RFC8693}}:
+A workload requesting a Txn-Token must provide the Transaction Token Service with proof of it's identity (client authentication), the purpose of the Txn-Token and any additional context relating to the transaction being performed. Most of these elements are provided by the OAuth 2.0 Token Exchange specification and the rest are defined as new parameters. This profile also defines a new token type URN `urn:ieft:params:oauth:token-type:txn-token` which is used by the requesting workload to identify that it is requesting the Txn-Token Response to contain a Txn-Token.
```
tulshi, line 395: Typo: it's -> its
Should we say "MUST provide the Txn-Token Service with proof"? (capitalize MUST)
Also, should we break this sentence into two parts: the required things (client-auth and purpose) and optional things (request context)? We should also drop the "also" from the last sentence (This profile also defines a new token type). We could even make that separate sub-section to clarify that we are defining a new token type.

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -365,24 +384,30 @@ The figure below {{figleaftxtokenbody}} shows a non-normative example of the JWT
...
 
-* The `audience` value MUST be set to the Trust Domain name
-* The `requested_token_type` value MUST be `urn:ietf:params:oauth:token-type:txn_token`
-* The `subject_token` value MUST be the external token received by the workload that authorized the call
-* The `subject_token_type` value MUST be present and indicate the type of the authorization token present in the `subject_token` parameter
+To request a Txn-Token the workload invokes the /token endpoint with the following parameters:
```
tulshi, line 397: Is the "/" required in the "/token"?

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -365,24 +384,30 @@ The figure below {{figleaftxtokenbody}} shows a non-normative example of the JWT
...
+* `subject_token_type` REQUIRED. The value MUST indicate the type of the token present in the `subject_token` parameter
 
-The following additional parameter MUST be present in a Txn-Token Request:
+The following additional parameters MAY be present in a Txn-Token Request:
 
-* A parameter named `rctx` , whose value is a JSON object. This object contains the request context, i.e. any information the Transaction Token Service needs to understand the context of the incoming request
+* `request_context` OPTIONAL. This parameter contains a base64url encoded JSON object which represents the context of this transaction. The parameter SHOULD be present and how the Transaction Token Service uses this parameter is out of scope for this specification.
+* `request_details` OPTIONAL. This parameter contains a base64url encoded JSON object which represents additional details of the transaction that MUST remain immutable throughout the processing of the transaction by multiple workloads.
+
+The requesting workload MUST authenticate it's identity to the Transaction Token Service. The exact client authentication mechanism used is outside the scope of this specification.
```
tulshi, line 410: it's -> its

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -391,20 +416,44 @@ POST /txn-token-service/token_endpoint HTTP 1.1
...
-&rctx=%7B%22param1%22%3A%22value1%22%2C%22param2%22%3A%22value2%22%2C%22ip_address%22%3A%2269.151.72.123%22%7D
+&request_context=eyAiaXBfYWRkcmVzcyI6ICIxMjcuMC4wLjEiLCAiY2xpZW50IjogIm1vYmlsZS1hcHAiLCAiY2xpZW50X3ZlcnNpb24iOiAidjExIiB9
 ~~~
 {: #figtxtokenrequest title="Example: Txn-Token Request"}
 
 
+## Txn-Token Request Processing
+When the Transaction Token Service receives a Txn-Token Request it MUST validate the requesting workload client authentication and determine if that workload is authorized to obtain the Txn-Tokens with the requested values. The authorization policy for determining such issuance is out of scope for this specification.
+
+Next, the Transaction Token Service MUST validate the `subject_token` and determine the value to specify as the `sub` of the issued Txn-Token. The Txn-Token Service MUST ensure the `sub` value is unique within the trust boundary defined by the `aud` claim.
```
tulshi, line 433: use "trust domain" instead of "trust boundary"?

draft-ietf-oauth-transaction-tokens.md
gffletch: I'm ok with that.

draft-ietf-oauth-transaction-tokens.md
gffletch: No, it's just my notation for identifying the endpoint. We could use "OAuth {{RFC6749}} token endpoint" to unique identify the endpoint.

draft-ietf-oauth-transaction-tokens.md
gffletch: @tulshi are you ok if I just change the MAY to MUST?  It is still possible for multiple Txn-Token issues to use a predetermined signing key.

I can add a sentence along the lines of "The signing key MUST be discoverable via the `iss` claim".

If we say that, do we need to define that discovery mechanism? or can we just refer to OAuth AS Metadata?

draft-ietf-oauth-transaction-tokens.md
gffletch: I took out the `iss` claim, and replaced `sub_id` with `sub`

draft-ietf-oauth-transaction-tokens.md
gffletch: I called out more clearly the optional parts from the required parts. Let me know what you think.

tulshi (approved):

PR #67, opened 2024-01-30
Clarify and align audience claim value format

### LINKED ISSUE #66: Trust domain/audience claim format URI or StringOrUri?
In the Terminology section, the Trust Domain value is being defined and referenced in the aud claim as universal resource identifier. It should be spelled Universal Resource Identifier and a reference to RFC3986 should be added.

However, later in the document, the aud claim value is specified as defined in RFC7519. This RFC defines the aud claim as "StringOrURI". The different passages should be aligned.

Furthermore, the examples for the aud claim values should be aligned accordingly. In section Txn-Token Request the POST request example to the token service uses:

`&audience=http%3A%2F%2Ftrust-domain.example`

If an URI is used, I suggest to use 'https' instead of 'http'.

Later the JWT body example of the Txn-Token just uses a string:

`"aud": "trust-domain.example",`

obfuscoder:
Resolve #66

### DIFF
--- a/draft-ietf-oauth-transaction-tokens.md
+++ b/draft-ietf-oauth-transaction-tokens.md
@@ -62,9 +62,9 @@ contributor:
 normative:
   RFC2119: # Keywords
   RFC8446: # TLS
-  RFC6749: #OAuth
-  RFC7519: #JWT
-  RFC7515: #JWS
+  RFC6749: # OAuth
+  RFC7519: # JWT
+  RFC7515: # JWS
   RFC8174: # Ambiguity in Keywords
   RFC8693: # OAuth 2.0 Token Exchange
   RFC8417: # Secure Event Token (SET)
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-transaction-tokens.md
```diff
@@ -263,7 +263,7 @@ Workload:
 : An independent computational unit that can autonomously receive and process invocations, and can generate invocations of other workloads. Examples of workloads include containerized microservices, monolithic services and infrastructure services such as managed databases.
 
 Trust Domain:
-: A virtually or physically separated network, which contains two or more workloads. The workloads within an Trust Domain may be invoked only through published interfaces. A Trust Domain must have an identifier that is used as the `aud` (audience) value in Txn-Tokens. The format of this identifier is a universal resource identifier. Each Trust Domain has exactly one Txn-Token Service.
+: A virtually or physically separated network, which contains two or more workloads. The workloads within an Trust Domain may be invoked only through published interfaces. A Trust Domain must have an identifier that is used as the `aud` (audience) value in Txn-Tokens. The format of this identifier is a StringOrURI as defined in {{RFC7519}}. Each Trust Domain has exactly one Txn-Token Service.
```
tulshi, line 266: specify "Section 2 of {{RFC7519}}"

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -347,7 +347,7 @@ The figure below {{figleaftxtokenbody}} shows a non-normative example of the JWT
 ~~~ json
 {
     "iat": "1686536226000",
-    "aud": "trust-domain.example",
+    "aud": "https://trust-domain.example",
```
tulshi, line 350: Do we need to have "https" or any protocol in a StringOrUri? It should be OK to not have a protocol prefix.

obfuscoder: True, if StringOrUri type should be defined, than leading https:// is not necessary. In this case I would align the other instances where this audience is used.

As you want to discuss it in the call today, I wait with the update of the PR until after the discussion.

(P.S: As much as I would like to join the regular call, the time slot is quite difficult for me due to other conflicting obligations.)

draft-ietf-oauth-transaction-tokens.md
gffletch: I'm not sure we want to require that "Each Trust Domain has `exactly one` Txn-Service". I'd prefer we remove this line entirely and be silent on the topic as there can be multiple deployment options.

PR #75, opened 2024-03-20
Section 6 - Cardinality of txn-token services

dhs-aws:
Changed MUST to MAY.  Cardinality of the txn-token service is 0..1, not exactly one.  

MAY not be the appropriate framing, I'm open to alternative text here to describe this more correctly.

### DIFF
--- a/draft-ietf-oauth-transaction-tokens.md
+++ b/draft-ietf-oauth-transaction-tokens.md
@@ -389,7 +389,7 @@ The figure below {{figleaftxtokenbody}} shows a non-normative example of the JWT
 # Txn-Token Service
 A Txn-Token Service defines a profile of the OAuth 2.0 Token Exchange {{RFC8693}} endpoint that can respond to Txn-Token issuance requests. This profile of the OAuth 2.0 Token Exchange {{RFC8693}} specification MUST be used to obtain Txn-Tokens. The unique properties of the Txn-Token requests and responses are described below. The Txn-Token Service MAY optionally support other OAuth 2.0 endpoints and features, but that is not a requirement for it to be a Txn-Token Service.
 
-Each Trust Domain MUST have exactly one logical Txn-Token Service.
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-transaction-tokens.md
```diff
@@ -389,7 +389,7 @@ The figure below {{figleaftxtokenbody}} shows a non-normative example of the JWT
 # Txn-Token Service
 A Txn-Token Service defines a profile of the OAuth 2.0 Token Exchange {{RFC8693}} endpoint that can respond to Txn-Token issuance requests. This profile of the OAuth 2.0 Token Exchange {{RFC8693}} specification MUST be used to obtain Txn-Tokens. The unique properties of the Txn-Token requests and responses are described below. The Txn-Token Service MAY optionally support other OAuth 2.0 endpoints and features, but that is not a requirement for it to be a Txn-Token Service.
 
-Each Trust Domain MUST have exactly one logical Txn-Token Service.
+Each Trust Domain MAY have exactly one logical Txn-Token Service.
```
tulshi, line 392: I'm not very familiar with the distinction here, but is there an optionality implied in the word "MAY"? I guess the intent is to say "one and only one logical Txn-Token Service"

dhs-aws: Perhaps the language should read, "...MUST have at most one logical Txn-Token session"

tulshi: Should we say "Each Trust Domain that uses Txn-Tokens MUST have exactly one logical Txn-Token Service."?

dhs-aws: That language works for me, @tulshi.

dhs-aws:
I've updated the PR with the language provided by @tulshi in two places where the previous language was found in the draft.

tulshi (approved):

PR #86, opened 2024-03-29
Addressed privacy changes identified by Yaron

tulshi:


gffletch (approved):
I'm ok with these changes. I think we could give some more advise around logging that could be useful to the industry.

tulshi:
I'll create a new issue to track that logging concern

Issue #76, opened 2024-03-26
Move 'aud' requirement in Trust Domain away from 'Terminology' section


tulshi:
Point from Yaron's [feedback email](https://mailarchive.ietf.org/arch/msg/oauth/qPxczdlhrkWbAu6eD8VmMmqRZj8/)

gffletch:
From Yaron:

> 4: the terminology section is not a good place for normative language, specifically around the "aud" claim.

gffletch:
Great point. I think I can move the key aspects into section 5.2

PR #89, opened 2024-03-29
Update draft-ietf-oauth-transaction-tokens.md

gffletch:
Addresses text regarding the `aud` claim throughout the document. Issue #76

### DIFF
--- a/draft-ietf-oauth-transaction-tokens.md
+++ b/draft-ietf-oauth-transaction-tokens.md
@@ -339,7 +339,7 @@ JWT claims as well as defines new claims. These claims are described below:
 : REQUIRED A unique transaction identifier as defined in Section 2.2 of {{RFC8417}}. When used in the transaction token, it identifies the entire call chain.
 
 `sub`:
-: REQUIRED A unique identifier for the subject within the context of the `aud` trust boundary. Unlike OpenID Connect, the `sub` claim is NOT associated with the `iss` claim.
+: REQUIRED A unique identifier for the subject within the context of the `aud` trust domain. Unlike OpenID Connect, the `sub` claim is NOT associated with the `iss` claim.
 
 `purp`:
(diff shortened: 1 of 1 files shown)


tulshi:
Although your changes sufficiently capture what is needed, what I liked about the previous draft is that it clearly identified that each Trust Domain MUST have a unique identifier, which is set as the `aud` value. I wonder if we can say that somewhere (outside of the Terminology section)

gffletch:
Ok, let me see if there is a good place to add that. I wonder if it will be easier to merge PR #90 first and then let me fix this one as there may be conflicts?

tulshi:
Makes sense to work on this after we merge #90

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -339,7 +339,7 @@ JWT claims as well as defines new claims. These claims are described below:
 : REQUIRED A unique transaction identifier as defined in Section 2.2 of {{RFC8417}}. When used in the transaction token, it identifies the entire call chain.
 
 `sub`:
-: REQUIRED A unique identifier for the subject as defined by the `aud` trust domain. Unlike OpenID Connect, the `sub` claim is NOT associated with the `iss` claim.
+: REQUIRED A unique identifier for the subject within the context of the `aud` trust boundary. Unlike OpenID Connect, the `sub` claim is NOT associated with the `iss` claim.
```
tulshi, line 342: Did you intend for "trust boundary" to be something separaet from "trust domain"? I'd rather not invoke an additional concept if we can avoid it.

gffletch: Yes, I agree. Let me fix that.

gffletch: Updated

tulshi (approved):

PR #90, opened 2024-04-12
specified language for self-signed JWTs as subject_tokens

tulshi:


draft-ietf-oauth-transaction-tokens.md
```diff
@@ -433,6 +437,19 @@ grant_type=urn%3Aietf%3Aparams%3Aoauth%3Agrant-type%3Atoken-exchange
 ~~~
 {: #figtxtokenrequest title="Example: Txn-Token Request"}
 
+### Subject Token Types
```
bc-pi, line 440: You might want to reconcile these with how Token Exchange (not particularity well) defines subject_token_type   https://datatracker.ietf.org/doc/html/rfc8693#section-2.1-4.14 and points to https://datatracker.ietf.org/doc/html/rfc8693#TokenTypeIdentifiers about identifiers

bc-pi: Oh and note Aaron's draft use of the `urn:ietf:params:oauth:token-type:id_token` URI for id tokens https://www.ietf.org/archive/id/draft-parecki-oauth-identity-assertion-authz-grant-00.html#section-5

gffletch: I agree with Brian. I think we need to make these URIs. If we want to explicitly support an rfc9068 token then we should define a URI for it. However, if the intent of the `rfc9068` value is to represent an access_token then we should use the value defined within RFC 8693.

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -433,6 +437,19 @@ grant_type=urn%3Aietf%3Aparams%3Aoauth%3Agrant-type%3Atoken-exchange
...
+The `subject_token_type` parameter in the Txn-Token Request MAY have one of the following values:
+
+rfc9068
+: The `subject_token` is a JWT format OAuth 2.0 Access Token as defined in {{RFC9068}}
+
+id-token
```
gffletch, line 446: RFC 8693 already defines a `subject_token_type` value for this which we should use. That said, `id_tokens` aren't usually authorization tokens and so I worry about explicitly calling it out here. The `id_token` can most definitely define the subject but then there is likely a need to pass in a different token containing the external authorization state. I'd rather we focus on AS generated access tokens and self-signed access tokens (recommended as RFC 9068).

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -433,6 +437,19 @@ grant_type=urn%3Aietf%3Aparams%3Aoauth%3Agrant-type%3Atoken-exchange
...
+: The `subject_token` is an OpenID Connect {{OpenIdConnect}} ID Token
+
+self-signed
+: The `subject_token` is a self-signed JWT issued by the requesting workload
+
+The Txn-Token Service MAY support other formats, which MAY be specified in the `subject_token_type` parameter
```
gffletch, line 452: Maybe add: "Note, as defined in RFC8693, the `subject_token_type` must be a URI.

gffletch (changes requested):
See inline comments.

draft-ietf-oauth-transaction-tokens.md
yaronf: True, and it means the current document should have an IANA subsection to define it.

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -433,6 +437,19 @@ grant_type=urn%3Aietf%3Aparams%3Aoauth%3Agrant-type%3Atoken-exchange
...
+The `subject_token_type` parameter in the Txn-Token Request MAY have one of the following values:
+
+rfc9068
+: The `subject_token` is a JWT format OAuth 2.0 Access Token as defined in {{RFC9068}}
+
+id-token
+: The `subject_token` is an OpenID Connect {{OpenIdConnect}} ID Token
+
+self-signed
+: The `subject_token` is a self-signed JWT issued by the requesting workload
```
yaronf, line 450: Other subject token types define the semantics of the token, and I think we should do that too. Maybe say (if this is true): 

> The `subject_token` is a self-signed OAuth access token issued by the requesting workload

yaronf (changes requested):

draft-ietf-oauth-transaction-tokens.md
bc-pi: It's really not an OAuth access token though (acknowledging that what actually constitutes an OAuth access token isn't totally clear but calling this an access token would be a stretch and unnecessarily confusing).

yaronf: @bc-pi OK, but I guess we should qualify this JWT to get interoperability. Should we say what claims are REQUIRED?

bc-pi: I don't know... I was only trying to politely point out that this thing is not an OAuth access token.

draft-ietf-oauth-transaction-tokens.md
tulshi: Thanks @bc-pi , @gffletch and @yaronf for your feedback here. I've updated the PR. Please take a look. thanks

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -438,18 +439,12 @@ grant_type=urn%3Aietf%3Aparams%3Aoauth%3Agrant-type%3Atoken-exchange
 {: #figtxtokenrequest title="Example: Txn-Token Request"}
 
 ### Subject Token Types
-The `subject_token_type` parameter in the Txn-Token Request MAY have one of the following values:
+The `subject_token_type` parameter value MUST be a URI {{RFC3986}}. It MAY be one of the subject token types described in Section 3 of OAuth 2.0 Token Exchange {{RFC8693}}, or it MAY be set to the value:
 
-rfc9068
-: The `subject_token` is a JWT format OAuth 2.0 Access Token as defined in {{RFC9068}}
+`urn:ietf:params:txn-token:token-type:self-signed`
```
yaronf, line 444: I'm pretty sure you need an IANA section to register `urn:ietf:params:txn-token`.

Hint: https://www.iana.org/assignments/params/params.xhtml

tulshi: I have fixed this (changed to the OAuth namespace and added an IANA registry section.

yaronf (approved):

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -143,6 +150,8 @@ Txn-Tokens are typically created when a workload is invoked using an endpoint th
 
 The Txn-Token Service responds to a successful invocation by generating a Txn-Token. The calling workload then uses the Txn-Token to authorize its calls to subsequent workloads. Subsequent workloads may obtain Txn-Tokens of their own.
 
+If the requesting service does not have an inbound token that it can use in its request to the Txn-Token Service, it generates a self-signed JWT and passes that in the request in place of the external authorization token.
+
```
PieterKas, line 154: Are there any RFCs that gives guidance on how to generate a self-signed JWT that we can reference? For example, can we reference RFC 7523?

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -406,7 +416,7 @@ To request a Txn-Token the workload invokes the OAuth 2.0 {{RFC6749}} token endp
 * `audience` REQUIRED. The value MUST be set to the Trust Domain name
 * `scope` REQUIRED. A space-delimited list of case-sensitive strings where the value(s) MUST represent the specific purpose or intent of the transaction.
 * `requested_token_type` REQUIRED. The value MUST be `urn:ietf:params:oauth:token-type:txn-token`
-* `subject_token` REQUIRED. The value MUST represent the subject of the transaction. This could be an OAuth access_token received by an API Gateway, a JWT assertion constructed by a workload initiating a transaction or a simple string value all identified by `subject_token_type`.
+* `subject_token` REQUIRED. The value MUST represent the subject of the transaction. This could be an inbound token received by an API Gateway, or a self-signed JWT constructed by a workload initiating a transaction, the type of which is identified by `subject_token_type`.
```
PieterKas, line 419: Perhaps add:

"The subject of this token will be the subject of the security token issued in response to the request." (RFC 8693 has similar terminology)

draft-ietf-oauth-transaction-tokens.md
tulshi: I didn't see any RFCs specific to self-signed JWTs. Referring to RFC7523 is probably not appropriate because of the additional constraints that RFC puts on the content of the JWT, which may not be what we intend here.

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -553,6 +570,13 @@ This specification registers the following claims defined in Section {{txn-token
 * Change Controller: IESG
 * Specification Document: Section {{txn-token-header}} of this specificaiton
 
+## OAuth URI Subregistry Contents
+
+* URN: urn:ietf:params:oauth:token-type:self_signed
+* Common Name: Token type for Self-signed JWT in lieu of an access token
```
obfuscoder, line 576: Instead of "in lieu" could you please use a more well-known English phrase?

draft-ietf-oauth-transaction-tokens.md
obfuscoder: The self-signed JWT should be short-living as well. So "iat" and "exp" should be included. there should probably some guidance on how short those self-sign JWTs should be valid. Could be same as for Txn-Tokens itself or even shorter as the self-signed JWTs here are meant to be used just once. "aud" and "sub" should be essential with "aud" either defining the trust domain (as described elsewhere in the spec) or narrowing down the audience even further to just the TTS by using the StringorURI of the TTS. Providing the "iss" claim could help the TTS to look up the necessary JWKS of the signing party.

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -433,6 +443,13 @@ grant_type=urn%3Aietf%3Aparams%3Aoauth%3Agrant-type%3Atoken-exchange
 ~~~
 {: #figtxtokenrequest title="Example: Txn-Token Request"}
 
+### Subject Token Types {#subject-token-types}
+The `subject_token_type` parameter value MUST be a URI {{RFC3986}}. It MAY be any one of the subject token types described in Section 3 of OAuth 2.0 Token Exchange {{RFC8693}} except the Refresh Token type (i.e., `urn:ietf:params:oauth:token-type:refresh_token`), or it MAY be set to the value:
+
+`urn:ietf:params:oauth:token-type:self_signed`
```
obfuscoder, line 449: Is defining and registering a new token type actually necessary? There is urn:ietf:params:oauth:token-type:jwt

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -433,6 +443,19 @@ grant_type=urn%3Aietf%3Aparams%3Aoauth%3Agrant-type%3Atoken-exchange
...
 
+## Subject Token Types {#subject-token-types}
+The `subject_token_type` parameter value MUST be a URI {{RFC3986}}. It MAY be any one of the subject token types described in Section 3 of OAuth 2.0 Token Exchange {{RFC8693}} except the Refresh Token type (i.e., `urn:ietf:params:oauth:token-type:refresh_token`), or it MAY be a self-signed JWT, as described below.
+
+The Txn-Token Service MAY support other token formats, which MAY be specified in the `subject_token_type` parameter. Any value used in this parameter MUST be a URI as specified in RFC 8693 {{RFC8693}}.
+
+### Self-Signed Subject Token Type {#self-signed-subject-token-type}
+A requester MAY use a self-signed JWT as a `subject_token` value. In that case, the requester MUST set the `subject_token_type` value to: `urn:ietf:params:oauth:token-type:self_signed`. This self-signed JWT MUST contain the following claims:
+
+* `iss`: The unique identifier of the requesting workload. The Txn-Token Service SHALL use this value in determiining the `req_wl` value in the Txn-Token issued in response to this request.
```
yaronf, line 454:
```suggestion
* `iss`: The unique identifier of the requesting workload. The Txn-Token Service SHALL use this value in determining the `req_wl` value in the Txn-Token issued in response to this request.
```

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -433,6 +443,19 @@ grant_type=urn%3Aietf%3Aparams%3Aoauth%3Agrant-type%3Atoken-exchange
...
+* `sub`: The subject for whom the Txn-Token is being requested. The Txn-Token Service SHALL use this value in determining the `sub` value in the Txn-Token issued in the response to this request.
```
yaronf, line 455: If we revert to `sub_id` for the main Txn Token, this one would also have to change.

yaronf (approved):

draft-ietf-oauth-transaction-tokens.md
bc-pi: > @bc-pi and @gffletch what are your thoughts on being more prescriptive on the format of the self-signed token? Can we leverage Section 3 of RFC 7523, or should we define our own? 

I wouldn't try to reference 7523 directly but maybe consider it (really just the client auth context) as partial inspiration for the claims that would make sense in this context of a self-signed JWT. Looks like @tulshi already started on a definition down in #self-signed-subject-token-type though and notably absent are `aud` and `exp`, which one would expect to see short of some compelling reason (that I'm not seeing).

tulshi: @bc-pi,  I considered adding `aud` and `exp` in the self-signed JWT, but here are the reasons why I decided not to:

- At a high-level, I'm thinking of the self-signed JWT in the request as a convenience rather than something that will have a lot of applicability outside of the communication between the requesting service and the TTS.
- We could set the `aud` value to be the TTS, but that is redundant. If we were to somehow hint about the `aud` value in the resulting TraT, that is determined by the TTS, so providing that in the request is inconsequential (and the requester has no control over it). It's possible that the same TTS has multiple trust domains, but there could be other ways for it to distinguish which trust domain it is acting on behalf of, such as a different token exchange endpoint URL.
- Similarly, the `exp` value could be used to indicate the expiration of the request itself, or it could be used as a hint to indicate the lifetime of the resulting TraT. Since we have already noted that a TTS may reject JWTs with an unreasonably old `iat` value, we do not need to specify an `exp` value. If we were to hint about the lifetime of the resulting TraT, I feel the requester would need to know a lot more about the transaction than a typical requester such as a general-purpose API gateway would know. I'd rather that the TTS determine the TraT's `exp` based on its policy and understanding of the actual transaction that is being requested.

PieterKas: @tulshi I am not so sure about not defining the aud value. Given everything that goes on in he world of token theft, having this extra constraint can help minimise the impact of token replay (stealing a token from one workload and replaying against another transaction token endpoint and so forth). Are there scenarios where we think the workload would not know which transaction token service it will be presenting its self-signed credential to?

I also recall that we had some constraint that a transaction token cannot be valid for longer than the original token. This raises the question on what the guidance is for expiry if the exp claim is not present as in this case. Section 2.3 is somewhat silent on this, but I think this is a decision for the Authorization service (perhaps something to be made clearer in section 2.3).

bc-pi: Claims like `exp`, `nbf`, `aud`, etc (also constructs like [explicit typing](https://www.rfc-editor.org/rfc/rfc8725.html#name-use-explicit-typing) fwiw) are about out the validity/security of the JWT in the intended context and also preventing misuse in unintended context. They have no bearing on whatever the output of the processes is. Any such bearing would need to be profiled and likely isn't what you'd want anyway. 

Note also that there are no normative processing requirements with iat.

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -542,7 +565,7 @@ Txn-Tokens SHOULD NOT be logged if they contain Personally Identifiable Informat
 
 # IANA Considerations {#IANA}
 
-This specification registers the following claims defined in Section {{txn-token-header}} to the OAuth Access Token Types Registry defined in {{RFC6749}}, and the following claims defined in Section {{txn-token-claims}} in the IANA JSON Web Token Claims Registry defined in {{RFC7519}}
+This specification registers the following claims defined in Section {{txn-token-header}} to the OAuth Access Token Types Registry defined in {{RFC6749}}, the following claim in Section {{subject-token-types}} to the "OAuth URI" subregistry of the "OAuth Parameters" {{IANA.OAuth.Parameters}} registry, and the following claims defined in Section {{txn-token-claims}} in the IANA JSON Web Token Claims Registry defined in {{RFC7519}}
```
bc-pi, line 568: most of these things are not claims

draft-ietf-oauth-transaction-tokens.md
tulshi: Thanks all, I've added the "aud" and "exp". Good points all (what was I thinking?). In order to minimize abuse of the self-signed JWT, I've specified that its lifetime should be just a few seconds. Since the Txn-token lifetime will exceed this, I've carved out that exception in "Txn-Token Lifetime" section (line 160)

draft-ietf-oauth-transaction-tokens.md
bc-pi: I realize much of this text is preexisting but it'd be good if the PR didn't introduce more misstatements

tulshi: Created separate PR for tracking this: https://github.com/oauth-wg/oauth-transaction-tokens/pull/92

bc-pi: #92 kinda misses the mark https://github.com/oauth-wg/oauth-transaction-tokens/pull/92#discussion_r1578619029 and the new text here is still problematic

obfuscoder:
@yaronf Regarding your comment about wrong approach: I mentioned something along those lines as well. I personally don't think that a new token type is actually necessary. A JWT is a good fit as it can carry the necessary data, but whether it should be signed by the workload (or what additional claims and associated checks should be incorporated) could be a decision by the individual TTS/trust domain.

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -433,6 +443,21 @@ grant_type=urn%3Aietf%3Aparams%3Aoauth%3Agrant-type%3Atoken-exchange
...
+
+### Self-Signed Subject Token Type {#self-signed-subject-token-type}
+A requester MAY use a self-signed JWT as a `subject_token` value. In that case, the requester MUST set the `subject_token_type` value to: `urn:ietf:params:oauth:token-type:self_signed`. This self-signed JWT MUST contain the following claims:
+
+* `iss`: The unique identifier of the requesting workload. The Txn-Token Service SHALL use this value in determining the `req_wl` value in the Txn-Token issued in response to this request.
+* `sub`: The subject for whom the Txn-Token is being requested. The Txn-Token Service SHALL use this value in determining the `sub` value in the Txn-Token issued in the response to this request.
+* `aud`: The unique identifier of the Txn-Token Service. The Txn-Token Service SHALL verify that this value matches its own unique identifier.
+* `iat`: The time at which the self-signed JWT was created. Note that the Txn-Token Service may reject self-signed tokens with an `iat` value that is unreasonably far in the past.
```
obfuscoder, line 457: or future?

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -433,6 +443,21 @@ grant_type=urn%3Aietf%3Aparams%3Aoauth%3Agrant-type%3Atoken-exchange
...
+* `exp`: The expiration time for the request. This should be a very short duration (order of seconds) in order to prevent any abuse of the JWT.
```
obfuscoder, line 458: "Expiration time for the request". It should be expiration time for the JWT. There is nothing said about checking this value by the TTS. I think we should be explicit here that TTS MUST validate this, but also clock skew needs to be addressed. We could also just refer the definition of the already registered claim name by RFC 7519

bc-pi: Kai is, of course, correct here that it's the expiration time of the JWT not of the request itself. And referring to existing definition of claims (not just exp) where possible and appropriate is likely preferable to (possibly erroneously)  restating/redefining them.

yaronf:
> @yaronf Regarding your comment about wrong approach: I mentioned something along those lines as well. I personally don't think that a new token type is actually necessary. A JWT is a good fit as it can carry the necessary data, but whether it should be signed by the workload (or what additional claims and associated checks should be incorporated) could be a decision by the individual TTS/trust domain.

@obfuscoder My point is that even the JWT structure/encoding is not needed. No need to sign anything, no need to encode stuff in base64 (and either of them adds potential security issues). All the semantics we need can be carried in a simple JSON object.

tulshi:
I tend to agree with Yaron, that perhaps self-signed JWTs aren't the best
way to express this. We can discuss this on the call tomorrow.

bc-pi:
Given the requirement that the TTS/AS always has to authenticate the identity of the requesting workload, conveying this info in an unsigned thing (the subject token could even contain just plain JSON) would probably be okay. Though it seems like the kind of thing that might lead to mistakes in deployment and resultant security issues.

PieterKas:
Letting the input be any kind of JSON structure feels problematic - not least because of interop. Using a JWT as the container and then putting constraints on it (i.e. a minimal set of supported claims) will help transaction token services in parsing and making some basic determinations about security. 

Self-signing has pros and cons - we can make it optional for those deployments that have a risk profile that would get additional assurances from it, but as Brian notes, signing is easy to get wrong.

bc-pi:
Well, you'd put some constraints on the JSON content too. I don't see how that be any more problematic from an interop perspective.

PieterKas:
As long as there are some constraints on the format and contents that everyone would expect. Why not just use a JWT at that point though (there are libraries that support it etc)?

draft-ietf-oauth-transaction-tokens.md
tulshi: This will certainly simplify the current PR, but would it prevent us in future from using the `urn:ietf:params:oauth:token-type:jwt` subject token type for other unanticipated purposes in the request?

draft-ietf-oauth-transaction-tokens.md
tulshi: fixed.

draft-ietf-oauth-transaction-tokens.md
tulshi: fixed.

draft-ietf-oauth-transaction-tokens.md
tulshi: I will address the broader issues with the nomenclature (header, media-type, etc.) in a separate PR (#92). For this PR, I have addressed the language regarding the token type identifier. @bc-pi  Please review if that part is OK.

tulshi:
> TBH I think we took the wrong approach here. Instead of defining a new token type and having to ensure it cannot be abused for anything else, we could have had a simple JSON object containing the relevant claims. The object would be carried as a separate, optional parameter. This would be so much simpler for developers and would reduce the attack surface by a ton.
> 
> With that said, I can live with the current solution.

@yaronf On [today's call](https://hackmd.io/@rpc-sec-wg/BJqJZLYW0) we discussed this topic, and while there are valid opinions on both approaches, for now I am keeping the request as a self-signed JWT unless we think this approach is problematic.

bc-pi (approved):

draft-ietf-oauth-transaction-tokens.md
obfuscoder: Good point. I also learned that token-type:jwt more or less defines the format of the token and not the purpose. Than again the name "token-type:self_signed" also seems to describe format rather than purpose.

How about registering a token type name based on the purpose and maybe avoid having "signed" in the type name while still defining the format as JWT with the needed attributes for interop and allow implementers to choose whether they want to require a signed JWT or allow a plain one. If implementers require a signed one, the additional attributes like "exp" could make sense to help prevent misuse/abuse. We could make those additional attributes OPTIONAL in the plain version and REQUIRED in the signed one.

draft-ietf-oauth-transaction-tokens.md
tulshi: Line 465 states that the `subject_token` is used to determine the `sub` of the Txn-Token. Is that good enough?

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -433,6 +443,13 @@ grant_type=urn%3Aietf%3Aparams%3Aoauth%3Agrant-type%3Atoken-exchange
 ~~~
 {: #figtxtokenrequest title="Example: Txn-Token Request"}
 
+### Subject Token Types {#subject-token-types}
+The `subject_token_type` parameter value MUST be a URI {{RFC3986}}. It MAY be one of the subject token types described in Section 3 of OAuth 2.0 Token Exchange {{RFC8693}} except the Refresh Token type (i.e., `urn:ietf:params:oauth:token-type:refresh_token`), or it MAY be set to the value:
```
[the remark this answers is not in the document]
tulshi: Wrote it as "MAY be any one of..."

draft-ietf-oauth-transaction-tokens.md
tulshi: Removed the additional phrase. The sentence makes sense without it.

gffletch (approved):
Let's go ahead and merge and then file issues if there is stuff we want to change.

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -143,13 +150,15 @@ Txn-Tokens are typically created when a workload is invoked using an endpoint th
...
+Txn-Tokens are expected to be short-lived (order of minutes, e.g., 5 minutes), and as a result MAY be used only for the expected duration of an external invocation. Except in the case where the request is made using a self-signed JWT, if the token or other credential presented to the Txn-Token service when requesting a Txn-Token has an expiration time, then the Txn-Token MUST NOT exceed the lifetime of the originally presented token or credential. If a long-running process such as an batch or offline task is involved, it can use a separate mechanism to perform the external invocation, but the resulting Txn-Token is still short-lived.
```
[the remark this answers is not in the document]
tulshi: Before we introduced self-signed JWTs in the request, the inbound tokens were expected to be longer lived. So this sentence said the Txn-Token MUST NOT exceed the lifetime of the inbound token. However, since self-signed JWTs are expected to be much more short-lived (as you mention), we want to carve out an exception for that. That is why I've made this change.

Issue #84, opened 2024-03-26
Possible confusion in IANA registry section


tulshi:
Yaron's [feedback email](https://mailarchive.ietf.org/arch/msg/oauth/qPxczdlhrkWbAu6eD8VmMmqRZj8/) states that:
"11.1: I think there is some confusion here. It is possibly useful to define this value (if we want to embed txn tokens within access tokens). But the "typ" header is a whole different thing, it needs to be a media type. See https://datatracker.ietf.org/doc/html/rfc8725#name-use-explicit-typing"

I'm not completely sure about this, so tracking this item

bc-pi:
https://hackmd.io/@rpc-sec-wg/SyAb2pxI6 from a while back had "https://www.ietf.org/archive/id/draft-ietf-oauth-transaction-tokens-00.html#name-iana-considerations Registration requests are off or missing or …."

PR #92, opened 2024-04-24
clarified fields versus claims and IANA registry name

tulshi:


draft-ietf-oauth-transaction-tokens.md
```diff
@@ -542,7 +542,7 @@ Txn-Tokens SHOULD NOT be logged if they contain Personally Identifiable Informat
 
 # IANA Considerations {#IANA}
 
-This specification registers the following claims defined in Section {{txn-token-header}} to the OAuth Access Token Types Registry defined in {{RFC6749}}, and the following claims defined in Section {{txn-token-claims}} in the IANA JSON Web Token Claims Registry defined in {{RFC7519}}
+This specification registers the following field defined in Section {{txn-token-header}} to the OAuth Access Token Types Registry defined in {{RFC6749}}, and the following claims defined in Section {{txn-token-claims}} in the IANA JSON Web Token Claims Registry defined in {{RFC7519}}
```
bc-pi, line 545: Lots to unpack here that goes beyond changing one word...

The referenced section `{{txn-token-header}}` is [JWT Header](https://www.ietf.org/archive/id/draft-ietf-oauth-transaction-tokens-01.html#name-jwt-header) which calls headers claims. Headers are not claims. Also the value of the `typ` header is supposed to be a media type so the `txn_token` value probably isn't quite right and the registration for it below certainly isn't right. Also the `urn:ieft:params:oauth:token-type:txn-token` URI is missing a registration and the [Txn-Token Request](https://www.ietf.org/archive/id/draft-ietf-oauth-transaction-tokens-01.html#name-txn-token-request) section that defines it has some major formatting issues. 

Some related reading, issues, and examples of these things being done in other specs follows: 

https://github.com/oauth-wg/oauth-transaction-tokens/issues/84
https://www.rfc-editor.org/rfc/rfc7515#section-4.1.9
https://www.rfc-editor.org/rfc/rfc7519#section-5.1
https://www.rfc-editor.org/rfc/rfc8725.html#name-use-explicit-typing
https://datatracker.ietf.org/doc/html/rfc9068#name-header
https://www.rfc-editor.org/rfc/rfc9449.html#name-dpop-proof-jwt-syntax
https://www.rfc-editor.org/rfc/rfc7519.html#section-10.2

tulshi: I'm pulling out the misuse of the "typ" Header Parameter as a separate issue so that it can be discussed on its own.

tulshi: https://github.com/oauth-wg/oauth-transaction-tokens/issues/97

bc-pi (approved):

PR #103, opened 2024-06-20
changing OAuth URI subregistry entry for Txn-Tokens to be txn_token

tulshi:
We currently use the URI: "urn:ietf:params:oauth:token-type:txn-token" to denote a Txn-Token in requests or in the token itself. However, if I look at other entries in the OAuth URI Subregistry here (https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml#uri), they use the "_token" style for the token types. We also use that style when the subject-token-type is "self_signed". So I'd like to propose that we use the URI: "urn:ietf:params:oauth:token-type:txn_token" for the transaction token URI.


PieterKas (approved):
Agreed with aligning to the common practices elsewhere.

gffletch (approved):

PR #104, opened 2024-06-21
draft02 files for submission. I have performed the IDNits check, and …

tulshi:
…the only issue is that there are non-ascii characters


gffletch (approved):
Not sure while the entire doc is shown as green. Approving.

tulshi:
closing this PR since we have moved to draft-03 now.

PR #122, opened 2024-08-05
added unsigned json subject token type

tulshi:


ashayraut (commented):
Please correct me if I am wrong but the github issue 121 asks "Add an outline on when self-signed may be useful". Which line here indicates "when" to use signed. Unsigned is in-expensive and low overhead for requesters so why would they ever go with Signed json, especially when in both signed or unsigned case, the Tx token may or may not consider the "sub" value in json to create Tx Token. The motivation (or mandate) to use signed json has to come from Tx token service. 

If Tx token service is going to use this value to find data about what claims to add or in other words if the claims are going to be determined based on sub value then requester MUST pass signed. If unsigned token is sent to Tx token that depends on signed token, then the claims might be missing from Tx token.

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -497,6 +500,14 @@ A requester MAY use a self-signed JWT as a `subject_token` value. In that case,
 
 The self-signed JWT MAY contain other claims.
 
+### Unsigned JSON Object Subject Token Type {#unsigned-json-subject-token-type}
+A requester MAY use an unsigned JSON object as a `subject_token` value. In that case, the requester MUST set the `subject_token_type` value to: `urn:ietf:params:oauth:token-type:unsigned_json`. The JSON object in the subject token MUST contain the following fields:
```
obfuscoder, line 504: How exactly is the  unsigned JSON encoded? Is it just a JSON string or is it at least base64urll encoded? We may need  to  mention this.

ashayraut: That's good callout to encode base64. Our language can be similar to what it says here in https://www.ietf.org/archive/id/draft-jones-json-web-token-02.html#rfc.section.7

tulshi: I believe a JWT is always Base64url encoded. It is either a JWS or a JWE, and those specs specify that it is encoded, e.g. in this section: https://datatracker.ietf.org/doc/html/rfc7515#section-5

tulshi:
> Please correct me if I am wrong but the github issue 121 asks "Add an outline on when self-signed may be useful". Which line here indicates "when" to use signed. Unsigned is in-expensive and low overhead for requesters so why would they ever go with Signed json, especially when in both signed or unsigned case, the Tx token may or may not consider the "sub" value in json to create Tx Token. The motivation (or mandate) to use signed json has to come from Tx token service.
> 
> If Tx token service is going to use this value to find data about what claims to add or in other words if the claims are going to be determined based on sub value then requester MUST pass signed. If unsigned token is sent to Tx token that depends on signed token, then the claims might be missing from Tx token.

The notes copy-pasted from the WG notes do say "when", but based on our recollection of the events during that session, the real ask was to clarify that one could use unsigned JSON objects. That is indicated in the action item (i.e. the last line in that issue description)

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -497,6 +500,14 @@ A requester MAY use a self-signed JWT as a `subject_token` value. In that case,
...
+
+* `sub`: The subject for whom the Txn-Token is being requested. The Txn-Token Service SHALL use this value in determining the `sub` value in the Txn-Token issued in the response to this request.
+* `exp`: The expiration time of the requested Txn-Token.
```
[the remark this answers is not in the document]
tulshi: Good point. I've added a line in the Txn-Token Request processing section to clarify.

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -482,9 +482,12 @@ grant_type=urn%3Aietf%3Aparams%3Aoauth%3Agrant-type%3Atoken-exchange
...
+The `subject_token_type` parameter value MUST be a URI {{RFC3986}}. It MAY be:
 
-The Txn-Token Service MAY support other token formats, which MAY be specified in the `subject_token_type` parameter. Any value used in this parameter MUST be a URI as specified in RFC 8693 {{RFC8693}}.
+* Any one of the subject token types described in Section 3 of OAuth 2.0 Token Exchange {{RFC8693}} except the Refresh Token type (i.e., `urn:ietf:params:oauth:token-type:refresh_token`).
+* A self-signed JWT, as described below.
+* An unsigned JSON object, as described below.
+* A custom URI agreed to between requesters and the Txn-Token Service. The Txn-Token Service MAY support other token formats, which MAY be specified in the `subject_token_type` parameter. Any value used in this parameter MUST be a URI as specified in RFC 8693 {{RFC8693}}.
```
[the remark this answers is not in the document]
tulshi: I'm OK with dropping this additional line. I thought someone had brought this up in our discussion at Vancouver, but I'll drop it for now.

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -482,9 +488,12 @@ grant_type=urn%3Aietf%3Aparams%3Aoauth%3Agrant-type%3Atoken-exchange
 {: #figtxtokenrequest title="Example: Txn-Token Request"}
 
 ## Subject Token Types {#subject-token-types}
-The `subject_token_type` parameter value MUST be a URI {{RFC3986}}. It MAY be any one of the subject token types described in Section 3 of OAuth 2.0 Token Exchange {{RFC8693}} except the Refresh Token type (i.e., `urn:ietf:params:oauth:token-type:refresh_token`), or it MAY be a self-signed JWT, as described below, or it MAY be a custom URI agreed to between requesters and the Txn-Token Service.
+The `subject_token_type` parameter value MUST be a URI {{RFC3986}}. It MAY be:
 
-The Txn-Token Service MAY support other token formats, which MAY be specified in the `subject_token_type` parameter. Any value used in this parameter MUST be a URI as specified in RFC 8693 {{RFC8693}}.
+* Any one of the subject token types described in Section 3 of OAuth 2.0 Token Exchange {{RFC8693}} except the Refresh Token type (i.e., `urn:ietf:params:oauth:token-type:refresh_token`).
+* A self-signed JWT, as described below.
+* An unsigned JSON object, as described below.
```
yaronf, line 495:
```suggestion
* A URN type name when the subject token is a self-signed JWT, as described below.
* A URN type name when the subject token is an unsigned JSON object, as described below.
```

Clearly the value of this field is not the self-signed JWT itself.

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -591,6 +610,9 @@ The `expires_in` and `scope` fields of the OAuth 2.0 Token Exchange specificatio
 ## Access Tokens
 When creating Txn-Tokens, the Txn-Token MUST NOT contain the Access Token presented to the external endpoint. If an Access Token is included in a Txn-Token, an attacker may extract the Access Token from the Txn-Token, and replay it to any Resource Server that can accept that Access Token. Txn-Token expiry does not protect against this attack since the Access Token may remain valid even after the Txn-Token has expired.
 
+## Subject Token Types {#sec-sub-token-types}
+A service requesting a Txn-Token SHOULD provide an incoming token if it has one that it used itself to authorize a caller, and if it directly correlates with the downstream call chain it needs the Txn-Token for. In the absence of an appropriate incoming token, the requesting service MAY use a self-signed JWT, an unsigned JSON object or any other format to represent the requester to the Txn-Token service.
```
yaronf, line 614: To represent the requester? Maybe "to express the properties it desires from the generated Txn-Token".

yaronf (changes requested):

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -454,7 +454,13 @@ To request a Txn-Token the workload invokes the OAuth 2.0 {{RFC6749}} token endp
...
 * `scope` REQUIRED. A space-delimited list of case-sensitive strings where the value(s) MUST represent the specific purpose or intent of the transaction.
 * `requested_token_type` REQUIRED. The value MUST be `urn:ietf:params:oauth:token-type:txn_token`
-* `subject_token` REQUIRED. The value MUST represent the subject of the transaction. This could be an inbound token received by an API Gateway, or a self-signed JWT constructed by a workload initiating a transaction, the type of which is identified by `subject_token_type`.
+* `subject_token` REQUIRED. The value MUST represent the subject of the transaction. This MAY be:
+  - An inbound token received by an API Gateway
+  - A self-signed JWT constructed by a workload initiating a transaction
+  - An unsigned JSON object constructed by a workload initiating a transaction
+  - Any other format that is understood by the Txn-Token Service
+
+  The type of `subject_token` field is identified by `subject_token_type`.
```
gffletch, line 463: add...  `the` -> The type of *the* `subject_token` ...

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -482,9 +488,12 @@ grant_type=urn%3Aietf%3Aparams%3Aoauth%3Agrant-type%3Atoken-exchange
...
+The `subject_token_type` parameter value MUST be a URI {{RFC3986}}. It MAY be:
 
-The Txn-Token Service MAY support other token formats, which MAY be specified in the `subject_token_type` parameter. Any value used in this parameter MUST be a URI as specified in RFC 8693 {{RFC8693}}.
+* Any one of the subject token types described in Section 3 of OAuth 2.0 Token Exchange {{RFC8693}} except the Refresh Token type (i.e., `urn:ietf:params:oauth:token-type:refresh_token`).
+* A URN type name when the subject token is a self-signed JWT, as described below.
+* A URN type name when the subject token is an unsigned JSON object, as described below.
+* A custom URI agreed to between requesters and the Txn-Token Service. The Txn-Token Service MAY support other token formats, which MAY be specified in the `subject_token_type` parameter.
```
gffletch, line 496: The "custom URI" could also be a URN correct?

draft-ietf-oauth-transaction-tokens.md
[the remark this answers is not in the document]
gffletch: I think the exp should the the validity of the unsigned JSON object to keep it semantically equivalent to the self-signed JWT "The expiration time for the JWT."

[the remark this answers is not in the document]
obfuscoder: @gffletch Although this would be more in line with JWTs, I'm not sure if this is useful. The unsigned JSON is typically constructed by the workload right before performing the request and it's usuallly a one-off. The server would rread the unsigned JSON and issue the Txn-Token after which it just thows away the unsigned JSON. It may help to detect large clock skews between workload and TTS though.

draft-ietf-oauth-transaction-tokens.md
obfuscoder: Yes JWTs are always base64 encoded, but we do not use an unsigned JWT here, but an unsigned JSON object. This is different.

draft-ietf-oauth-transaction-tokens.md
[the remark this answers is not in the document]
gffletch: I just don't like changing the semantics of the `exp` field between different token types. What about describing the `exp` value as...

`exp` : The expiry time of the unsigned JSON object which the TTS MAY use as input to determining the life-time of the Txn-token.

If we need a way for the requesting workload to request an expiry time for the Txn-token then we should allow that for all flows and not just this one.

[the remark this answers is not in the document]
tulshi: Agreed, used your proposed language.

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -497,17 +507,28 @@ A requester MAY use a self-signed JWT as a `subject_token` value. In that case,
...
 
 Next, the Transaction Token Service MUST validate the `subject_token` and determine the value to specify as the `sub` of the issued Txn-Token. The Txn-Token Service MUST ensure the `sub` value is unique within the trust domain defined by the `aud` claim.
 
 The Transaction Token Service MUST set the `iat` claim to the time of issuance of the Txn-Token.
+
 The Transaction Token Service MUST set the `aud` claim to an identifier representing the Trust Domain of the Transaction Token Service. If the Transaction Token Service supports multiple trust domains, then it MUST determine the correct `aud` value for this request.
-The Transaction Token Service MUST set the `exp` claim to the expiry time of the Txn-Token.
+
+The Transaction Token Service MUST set the `exp` claim to the expiry time of the Txn-Token. The Txn-Token Service MAY consider any `exp` value present in the `subject_token` field of the Txn-Token Request in determining the `exp` value of the resulting Txn-Token.
```
gffletch, line 527: Maybe change "`subject_token` field" to "`subject_token` parameter" ?

gffletch (approved):
Just a few very minor comments

draft-ietf-oauth-transaction-tokens.md
tulshi: done.

PR #123, opened 2024-08-26
moved txn field usage information to Security Considerations section

tulshi:


draft-ietf-oauth-transaction-tokens.md
```diff
@@ -600,6 +600,9 @@ Validation of a replacement Txn-Token, as well as any Txn-Token, is critical to
 ## Scope and Purpose processing
 The authorization model within a trust domain boundary is most often quite different from the authorization model (e.g. OAuth scopes) used with client external to the trust domain. This makes managing unintentional scope increase a critical aspect of the Transaction Token Service. The TTS MUST ensure that the requested purpose (`scope`) of the Txn-Token is equal or less than the scope(s) identified in the `subject_token`. This is also true of requesting a replacement Txn-Token. The TTS MUST ensure there is not unintentional increase in authorization scope.
 
+## Identifying Call Chains
+A Txn-Token typically represents one call-chain following an external invocation. It is thus useful to identify the call-chain using the Txn-Token. The `txn` claim in the Txn-Token may be used for this purpose. When used in the Txn-Token, it identifies the entire call chain. It is strongly RECOMMENDED to provide an identifier unique within the trust domain.
```
ashayraut, line 604: Txn identifier is useful for auditing and debugging. We have to make sure TTS logs Tx identifier on its end and then every service in call chain also logs Txn identifier, so in case, a Tx token validating party reports some problem like missing claim or security auditors want to verify why some data was allowed/denied access, then its useful to go back to TTS and verify what happened exactly using token issuance using the Tx identifier.

We should include TTS logging Txn identifier.

gffletch: I would recommend "following an *initial* invocation" or "following an *initiating* invocation" so that we don't bias this toward external calls vs inside-the-perimeter invocations.

gffletch: What about...

A Txn-token typically represents the call-chain of workloads necessary to complete a logical function initiated by an external or internal workload. The `txn` claim in the Txn-token provides a unique identifier that when logged by the TTS and each subsequent workload can provide both discovery and auditability of successful and failed transactions. It is therefore strongly RECOMMENDED to use an identifier, unique within the trust domain, for the `txn` value.

gffletch (approved):

PR #125, opened 2024-09-05
clarified logging recommendation

tulshi:


ashayraut (approved):
Looks good to me.

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -635,7 +635,7 @@ and if so their values need to be obfuscated. For example, originating IP addres
 protected through some obfuscation method (e.g. salted SHA256).
 
 ## Logging
-Txn-Tokens SHOULD NOT be logged if they contain Personally Identifiable Information (PII). What constitutes PII depends upon the use case, but in some cases even an email address (which could be a `sub` value) can be protected PII, which should not be logged.
+Txn-Tokens should not be logged. This is in order to prevent abuse of tokens via log files. The content from the Txn-Tokens may be logged after the signature has been removed. If the Txn-Token contains PII, then care should be taken in logging the content of the Txn-Token so that the PII does not get logged.
```
gffletch, line 638: Removing the signature is one way to make the token not replayable, but it's not the only way to address the logging concerns. For instance, select fields could be extracted from the token and logged. 

Complete Txn-Tokens must not be logged verbatim due to the potential replay risk of tokens extracted from the log files. Additionally, Txn-Tokens may contain PII which should be anonymized or removed before logging the values. There are multiple ways to ensure the logged data can not be replayed for example logging a Txn-Token that does not contain any PII after removing the signature from the base64url encoded token.

Or something like that?

PieterKas: I think there are multiple forms of abuse, including replay and information leakage.

I worry about schemes that strip the signature - it limits replay risk (assuming the signature is checked.... if it is not, there are bigger problems), but removes the integrity protection allowing an attacker to manipulate the contents of the token which may make forensic analysis a bit more tricky, so stripping the signature is safe as long as the log file's integrity is protected.

Building on the existing text, something like:

Txn-Tokens should not be logged by microservices. This is in order to prevent replay of tokens or leakage of PII or other sensitive information via log files. A hash of the Txn-Token may be logged to allow for correlation with the log files of the Txn-Token Service that records issued tokens. Alternatively the JWS payload of a Txn-Token may be logged after the signature has been removed. If the Txn-Token contains PII, then care should be taken in logging the content of the Txn-Token so that the PII does not get logged.

tulshi: I've incorporated some of both your comments here.

gffletch (approved):

obfuscoder:
Is there a reason why you did not use uppercase terms like MUST NOT or SHOULD? Because it's supposed to be informative?

tulshi:
Yes, the section is non-normative.

gffletch (approved):

Issue #127, opened 2024-09-23
Clarify Transaction Token Context


PieterKas:
In section 2.2.1 the context that may be included in a transaction token includes "The external authorization token (e.g., the OAuth access token)". To ensure this is not interpreted to mean that an access token is included in a transaction token, I would suggest the following additions, in line with section 9.3 in Security Considerations:

"A reference to the external authorization token (e.g., the OAuth access token), including scopes or claims included in the authorization token, but not the unmodified authorization token (see Security Considerations, Section 9.3)"

gffletch:
Pieter, in that section of the spec we are describing what needs to be sent to the Transaction token service, I think in that case the full access token should be passed. However, when generating the request context, we want to ensure that the full access token is not included. Did I get my spec sections wrong?

gffletch:
Maybe instead we shouldn't say "This context MAY include:" but rather ... "The information provided to the Txn-Token Service MAY include" to not confuse readers who might think this data should be included in the request context.

PieterKas:
Agreed - I missed that this was for initial token creation. I did create a PR with your suggested text.

gffletch:
PR approved

PR #137, opened 2024-09-27
Clarify Transaction Token Context

PieterKas:
See issue #127


draft-ietf-oauth-transaction-tokens.md
```diff
@@ -156,7 +156,7 @@ Txn-Tokens are short-lived, signed JWTs {{RFC7519}} that assert the identity of
 ### Initial Creation
 Txn-Tokens are typically created when a workload is invoked using an endpoint that is externally visible, and is authorized using a separate mechanism, such as an OAuth {{RFC6749}} access token or an OpenID Connect {{OpenIdConnect}} ID token. This workload then performs an OAuth 2.0 Token Exchange {{RFC8693}} to obtain a Txn-Token. To do this, it invokes a special Token Service (the Txn-Token Service) and provides context that is sufficient for it to generate a Txn-Token. This context MAY include:
 
-* The external authorization token (e.g., the OAuth access token)
+* A reference to the external authorization token (e.g., the OAuth access token). This may inlcude a hash of the authorization token and may including scopes or claims included in the authorization token. To minimise the risk of token theft, It MUST NOT include the unmodified authorization token (see Security Considerations, Section 9.3).
```
gffletch, line 159: In this context of the spec... what needs to be sent to the Txn-Token Service, I believe we want to send the full access token. If not, we should have a longer discussion about what should be sent. My expectation has been that the TTS would validate the access_token for it's authenticity before evaluating the TTS authorization policy to determine whether a TraT should be issued.

Issue #129, opened 2024-09-23
Trust Domain definition


PieterKas:
The definition of a Trust Domain feels very specific to network segmentation. Is there a industry level Trust Domain definition we can use instead? For example, should we use a definition that defines the trust domain in terms of the shared policies that apply? Something like:

"A trust domain refers to a collection of systems, applications, or workloads that share a common security policy".

PR #138, opened 2024-09-27
Trust domain clarification

PieterKas:
See issue #129

Question for reviewers. 

1. Do we need the final sentence on how a workload is invoked? If the sentence was not there would it change how we think about a trust domain?

2. Do we need the second sentence as an example? Should we add more examples (e.g. trust domains can be defined as all the applications or workloads that recognise a specific issuer).


gffletch (approved):
I'm good with this

Issue #134, opened 2024-09-23
Clarify why the "aud" claim remains unchanged


PieterKas:
Good practice is to explain why something is a MUST. 

Suggestion:

From: The value of the aud claim MUST remain unchanged in a replacement Txn-Token.

To: The value of the aud claim MUST remain unchanged in a replacement Txn-Token to prevent the Txn-Token from being accepted outside it's current Trust Domain.

Alternatively, we cann add security considerations for the draft as well.

gffletch:
I'm good with that text.

tulshi:
I think the current PR is making the change in the wrong place (unless I'm horribly mistaken)

PR #141, opened 2024-09-27
Additional detail on 'aud' claim

PieterKas:
See issue #134

### DIFF
--- a/draft-ietf-oauth-transaction-tokens.md
+++ b/draft-ietf-oauth-transaction-tokens.md
@@ -342,7 +342,7 @@ JWT claims as well as defines new claims. These claims are described below:
 : REQUIRED The issued at time of the Txn-Token as defined in {{RFC7519}}
 
 `aud`:
-: REQUIRED This claim, defined in {{RFC7519}}, identifies the trust domain in which the Txn-Token is valid.  This identifier MUST uniquely identify the trust domain to prevent the Txn-Token from being accepted outside it's current Trust Domain.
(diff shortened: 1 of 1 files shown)


gffletch (approved):
Looks good

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -451,8 +451,8 @@ A workload requesting a Txn-Token must provide the Transaction Token Service wit
 
 To request a Txn-Token the workload invokes the OAuth 2.0 {{RFC6749}} token endpoint with the following parameters:
 
-* `grant_type` REQUIRED. The value MUST be set to `urn:ietf:params:oauth:grant-type:token-exchange`
-* `audience` REQUIRED. The value MUST be set to the Trust Domain name
+* `grant_type` REQUIRED. The value MUST be set to `urn:ietf:params:oauth:grant-type:token-exchange`.
+* `audience` REQUIRED. The value of the aud claim MUST remain unchanged in a replacement Txn-Token to prevent the Txn-Token from being accepted outside it's current Trust Domain.
```
tulshi, line 455: This is in the "Txn-Token Request" section. Do we need to talk about replacement here, or should we have that language in the "Creating Replacement Txn-Tokens" section? We perhaps already have this there. Also, if you delete the "value MUST be set to the Trust Domain name", then we don't have that constraint listed anywhere else. We should make sure we have that specified somewhere if we want to remove it from here.

PieterKas: When the aud claim is initially defined, it is constrained to the trust domain. Agreed this is in the wrong section. Thanks for catching that. Working on an update,

tulshi (approved):

PR #149, opened 2024-11-14
Add Batch token for asynchronous long running workloads

ashayraut:
Batch token was discussed here at https://github.com/oauth-wg/oauth-transaction-tokens/issues/111. The PR is first draft to add batch token in RFC.

### DIFF
--- a/draft-ietf-oauth-transaction-tokens-03.md
+++ b/draft-ietf-oauth-transaction-tokens-03.md
@@ -168,8 +168,14 @@ A service within a call chain may choose to replace the Txn-Token. This can typi
 
 To get a replacement Txn-Token, a service will request a new Txn-Token from the Txn-Token Service and provide the current Txn-Token and other parameters in the request. The Txn-Token service must exercise caution in what kinds of replacement requests it supports so as to not negate the entire value of Txn-Tokens.
 
+## Batch Txn-Tokens
+
+A service within a call chain may start asynchronous flows by invoking a workflow, or sending a message in queue or starting a batch job. Asynchronous flows may take minutes, hours or days to complete depending on the business use case. The service cannot pass Txn-Token to downstream as the asynchronous flow could outlive the lifetime of the Txn-Token. In such cases, the service can call Txn-Token Service to exchange a valid Txn-Token for a long lived Batch Txn-Token. Then use the Batch Txn-Token to invoke downstream asynchronous workload. At later point of time, when asynchronous workflow requires to call an internal service, resembling a synchronous flow, the workload can use the Batch Txn-Token to call Txn-Token service and request a valid short lived Txn-Token. The workload will then use the Txn-Token to call another internal service. A Batch Txn-Token MUST be encrypted and only Txn-Token Service MUST be able to decrypt a Batch Txn-Token. A validating party will not be able to decrypt the Batch Txn-Token. During exchange of a Batch Txn-Token to Txn-Token, Txn-Token service MUST only add those claims back which are still valid at that issuance time. 
+
+In order for Txn-Token Service to be able to issue a valid token in exchange for a Batch Txn-Token, it must be store the claims in the Txn-Token that was first used to generate a Batch Txn-Token. Txn-Token Service can choose to store claims in the Batch Txn-Token itself or separately in a persisent store and add a unique identifier in the Batch Txn-Token that points to the claims in the storage.
+
 ## Txn-Token Lifetime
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-transaction-tokens-03.md
```diff
@@ -168,8 +168,14 @@ A service within a call chain may choose to replace the Txn-Token. This can typi
...
 
+## Batch Txn-Tokens
+
+A service within a call chain may start asynchronous flows by invoking a workflow, or sending a message in queue or starting a batch job. Asynchronous flows may take minutes, hours or days to complete depending on the business use case. The service cannot pass Txn-Token to downstream as the asynchronous flow could outlive the lifetime of the Txn-Token. In such cases, the service can call Txn-Token Service to exchange a valid Txn-Token for a long lived Batch Txn-Token. Then use the Batch Txn-Token to invoke downstream asynchronous workload. At later point of time, when asynchronous workflow requires to call an internal service, resembling a synchronous flow, the workload can use the Batch Txn-Token to call Txn-Token service and request a valid short lived Txn-Token. The workload will then use the Txn-Token to call another internal service. A Batch Txn-Token MUST be encrypted and only Txn-Token Service MUST be able to decrypt a Batch Txn-Token. A validating party will not be able to decrypt the Batch Txn-Token. During exchange of a Batch Txn-Token to Txn-Token, Txn-Token service MUST only add those claims back which are still valid at that issuance time.
```
pdvrieze, line 173: This would effectively function as a refresh token. Having an opaque (random) value (not encrypted) would be sufficient (encrypted token could also work).

It would be important to require that the service must authenticate itself to acquire the batch token, and that the batch token should be bound to the originating service. For a long-lived token it would also be important to define invalidation semantics (e.g. based upon the `Txn-Id`).

An alternative approach would be to allow Txn-Token refreshing with service identity.

draft-ietf-oauth-transaction-tokens-03.md
```diff
@@ -278,6 +284,81 @@ In the diagram above, steps 1-5 are the same as in {{basic-flow}}
...
+~~~
+{: #fig-arch-async-flow title="Batch Txn-Token Flow"}
+
+In the diagram above, steps 1-5 are the same as in {{basic-flow}}
+
+{:start="6"}
+
+6. An intermediate service determines that it needs to start a workflow. Once workflow is started successfully, then it can respond back to incoming request. The workflow will then can run for longer duration. The intermediate service needs to obtain a Batch Txn-Token to pass to the workflow as the workflow execution will outlive Txn-Token lifetime. It requests a Batch Txn-Token from the Txn-Token Service. It passes the incoming Txn-Token in the request, along with any additional context it needs to send the Txn-Token Service.
+7. The Txn-Token Service responds with a Batch Txn-Token
+8. The service that requested the Batch Txn-Token uses that Batch Txn-Token for invoking workflow
```
pdvrieze, line 353: Would it not be the responsibility of the workflow system to acquire/exchange the batch token? If the service creating the batch token is not the same as the one using it to create further tokens, there may be opportunities for breaches. At the very least the workflow/receiver would not be able to verify the batch token validity.

pdvrieze (commented):
There are some further considerations, including requiring service/client authentication, revocation and (potentially) binding the batch token to be created/used by the same service.

draft-ietf-oauth-transaction-tokens-03.md
obfuscoder: Is this PR planned to make it into the final draft of the spec or will it be paused until after releasing the first draft?

In regards to a Batch Token compared to a Refresh Token: Is it planned to allow the Batch Token to be exchanged to a new Txn-Token multiple times or do we allow to exchange the Batch Token only once similar? If it is supposed to be used only once, it would have more similarities with an Authorization Code instead of a Refresh Token (although an Authorization Code is supposed to be short-lived).

Client/Service authentication is indeed something which needs to be addressed. As this is currently lacking, basically any workload can exchange a Batch Token back to a Txn-Token at any given time with the help of the Txn-Token Service. Essentially, a workload can exchange a Txn-Token to a Batch Token send this Batch Token along to other entities (e.g. Kafka event queues or store in a database) and any other entity having access to that Batch Token (e.g. Kafka event processors) can get an updated Txn-Token based on the Batch Token again. When issuing the Batch Token, do we want to limit the clients which can later exchange the Batch Token back to an Txn-Token or do we trust that the implementers will reasonably protect access to Batch Tokens? In case of limiting access within this spec, we might need to encode the audience(s) which are allowed to exchange the Batch Token later.

How long will Batch Tokens be valid? I assume at maximum for how long the keys are available to the Txn-Token Service to decrypt the content of the Batch Token.

Issue #164, opened 2025-03-25
Expanding scope of permitted actions


dhs-BI:
Section 7.5.1 states, "SHOULD NOT enable modification to asserted values that expand the scope of permitted actions".  This seems overly permissive, is there a reason the sentence does not begin with "MUST NOT"?

Section 7.5.2 appears to have what I would consider the correct verbiage in the final sentence, "[...] MUST NOT increase the authorization surface [...]".

pdvrieze:
There is actually a use case that hits this. From a minimal permissions perspective it should be possible for the delegate service to perform actions that its caller cannot. However in such cases this client scoped permission can still be specified/determined on the creation of the transaction (independently it may or may not be a good idea to hide these scopes from the token).

PieterKas:
We should not add more scope than the original Access Token. Add clarification.

pdvrieze:
@PieterKas What I am thinking of is (in the context of services called by a process engine) the following:
- Take 4 services A, B, C, and D (and call sequences: A→B, B→C, C→D) and corresponding tokens
  - A→B : T1
  - B→C : T2
  - C→D: T3
- A is the initiating service that starts the transaction (for example a process engine).
- Updates to `D` are sensitive, and `B` should not be trusted to directly invoke `D` (it has no need to either - minimal permissions)
- In such a case it would be an option to have the following permission model
- T1:
  - permits A access to B with a given scope (client and audience could be bound)
  - permits B to exchange T1 for T2 (with the pre-specified permissions):
    - permit direct access by B to C for the specified scope
    - permit exchange by C for a new token (T3):
      - C is permitted to access D (with given scope)
- T2:
  - permits B access to C with a given scope (client and audience could be bound)
  - permits C to exchange T2 for T3:
    - C is permitted to access D (with given scope)
- T3:
  - permits C to access D

If we consider that token exchange is designed to alter the security context and permissions (including client bound tokens - e.g. DPoP) then it makes sense to allow for specifying (in some way this is therefore still narrowing) permissions for a service lower in the invocation chain. As to context, imagine that service B is an external service, or closed source software with limited trustworthiness (Service C can perform the appropriate checks before invoking D).

PieterKas:
@pdvrieze we should distinguish between access required to complete the transaction (all of which was consented to up-front on transaction initiation) and what individual workloads may be permitted to do while executing the transaction. 

[...]

PR #173, opened 2025-04-28
Strengthen prohibition on expanding scope

PieterKas:
Update to be consistent in prohibition of extending the scope of an Access Token. See issue #164


tulshi (approved):

gffletch:
I'm happy to approve it. I seem to remember there being some concern that enforcing `MUST NOT` may not always be possible and hence we should use `SHOULD NOT`. That said, we can change it and see if anyone has comments or concerns.

PR #185, opened 2025-08-05
renamed microservice to workload

tulshi:


draft-ietf-oauth-transaction-tokens.md
```diff
@@ -190,8 +190,7 @@ Txn-Tokens help prevent spurious invocations by ensuring that a workload receivi
              4 v   │ 6
           +--------------+
           │              │
-          │   Internal   │
```
gffletch, line 193: Do we want to drop the concept of "Internal"? I'm ok but I feel it does help with the concept of restricting transaction tokens to a single trust domain. Thoughts?

Issue #192, opened 2025-08-26
Be more specific on Authorization Context


PieterKas:
WGLC Feedback from @bc-pi 

https://datatracker.ietf.org/doc/html/draft-ietf-oauth-transaction-tokens-06#section-4-1.12.1 
Authorization Context, to me anyway, seems like too general of a concept/thing to "define" as a "JSON object".

PR #237, opened 2025-08-28
Proposed change to definition of authorization context

PieterKas:
Proposed response to issue #192

@bc-pi

### DIFF
--- a/draft-ietf-oauth-transaction-tokens.md
+++ b/draft-ietf-oauth-transaction-tokens.md
@@ -741,6 +741,10 @@ The authors would like to thank the contributors and the OAuth working group mem
 {: numbered="false"}
 [[ To be removed from final specification ]]
 
+## Since Draft 06
+{:numbered="false"}
+* Remove definition of Authorization Context [Be more specific on Authorization Context](https://github.com/oauth-wg/oauth-transaction-tokens/issues/192)
+
 ## Since Draft 05
 {:numbered="false"}
 * [Strengthened prohibition on expanding TraT scope](https://github.com/oauth-wg/oauth-transaction-tokens/pull/173)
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-transaction-tokens.md
```diff
@@ -303,7 +303,7 @@ Transaction Token (Txn-Token):
 : A signed JWT with a short lifetime, providing immutable information about the user or workload, certain parameters of the call, and specific contextual attributes of the call. The Txn-Token is used to authorize subsequent calls in the call chain.
 
 Authorization Context:
```
bc-pi, line 305:
```suggestion
```

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -303,7 +303,7 @@ Transaction Token (Txn-Token):
...
-: A JSON object containing a set of claims that represent the immutable context of a call chain.
+: The information that represent the immutable context of a call chain. This information MAY be expressed as a JSON object containing a set of claims.
```
bc-pi, line 306:
```suggestion
```

draft-ietf-oauth-transaction-tokens.md
bc-pi: the words authorization context appear three other times in the document, before this, with lower case a and c. 
this definition doesn't honestly make sense to me but before tying to figure that out i'd suggest that having it in the terminology isn't even needed

bc-pi (changes requested):
maybe just remove it

PieterKas:
Agreed that given that the definition is not used, having it formally defined seems unnecessary. 

also added document history.

bc-pi (approved):

Issue #194, opened 2025-08-26
Reconsider 'purp' claim scope


PieterKas:
WGLC feedback from @bc-pi 

https://datatracker.ietf.org/doc/html/draft-ietf-oauth-transaction-tokens-06#section-5.2-2.14.1 and https://datatracker.ietf.org/doc/html/draft-ietf-oauth-transaction-tokens-06#name-purpose-claim 
Not sure how to articulate this but the 'purp' claim feels simultaneously too narrow and too general to be meaningfully useful. Particularly for being required. And with the tctx claim also.

PieterKas:
My mental model for the "purp" is that it is a kind of "fine grained scope" and allows the transaction server to communicate that access granted is smaller or equal to "scope". It is seperate from transaction context. It is a mechanism to downscope "scope".

Not sure that helps @gffletch and @tulshi should weigh in here as well.

dagdagdag83:
I share the sentiment of Brian, I never quite understood the purp claim. It just smells bad. It feels redundant somehow to scope, and I never quite understood if it's more specific or more general. And I've read the draft many times.

I have the feeling that purp was made to give a different claim, so that the scope(s) of any tokens in the flow doesn't get directly confused with a scope claim in transaction tokens. But it just feels different, but also to similar. Does that make sense?

bc-pi:
> My mental model for the "purp" is that it is a kind of "fine grained scope" and allows the transaction server to communicate that access granted is smaller or equal to "scope". It is seperate from transaction context. It is a mechanism to downscope "scope".
> 
> Not sure that helps [@gffletch](https://github.com/gffletch) and [@tulshi](https://github.com/tulshi) should weigh in here as well.

It helps confirm to me what was said earlier:

>  ... scope can still be used to convey scope, even if it's inside and/or more granular and/or different.

tulshi:
I feel having a different name (i.e., "purp"), instead of the "scope" helps distinguish between the OAuth scope of an inbound token versus the similarly useful field of the TraT.

bc-pi:
I see I'd previously had a typo/misspelling where I wrote the word "compiling" where I meant "compelling". I don't find the rational for this "purp" claim to be compelling.

PieterKas:
@bc-pi - in the absence of a "purp" claim, how would you suggest  scope is further reduced from the original scope?

tulshi:
[...]

Issue #178, opened 2025-06-20
Consider replacing Scope with Purpose


PieterKas:
From the mailing list:

https://mailarchive.ietf.org/arch/msg/oauth/tOA3lQDoiij0sPi7JI2nFDNOPbA/

[OAUTH-WG] Tx-Token Purpose vs Scope
Joe DeCock <joe@duendesoftware.com> Thu, 19 June 2025 22:44 UTC[Show header](https://mailarchive.ietf.org/arch/msg/oauth/tOA3lQDoiij0sPi7JI2nFDNOPbA/#)

Hello OAuth working group,

I've been reading the transaction tokens draft with great interest. I think
it's solving an important problem and largely I like what I'm seeing in the
draft. However, I think that the way that scope and purpose are
differentiated can be improved to make the specification easier to
understand, easier to implement, and improve interoperability.

The way that the purpose is described and contrasted with OAuth scopes in
section 5.2.1 makes good sense to me.

The purp claim captures the exact purpose of this particular transaction.
> This is often much narrower than a scope value issued to an external
> client. This is due to the fact that in most cases, the authorization model
> within the Trust Domain is quite different than the authorization model
> used with clients external to the Trust Domain. To that end, it is
> intentional to separate the concept of scope (often fairly coarse-grained)
> used with external clients from the purpose of the transaction used within
> the Trust Domain. How a given deployment represents the authorization model
> within the Trust Domain is out of scope for this specification.

But in the txn-token request (Section 7.1), the purpose of the transaction
token is sent as the scope parameter. I think that has the potential to be
confusing, because the two concepts are meant to be distinct. In my
experience, non-identity experts often struggle with the existing concepts
of scopes, claims, and roles. Reusing the scope parameter with purpose
semantics is likely going to be a cause for confusion.

Section 9.6 muddies the distinction between scope and purpose further:

The TTS MUST ensure that the requested purpose (scope) of the Txn-Token is
> equal or less than the scope(s) identified in the subject_token. This is
> also true of requesting a replacement Txn-Token. The TTS MUST ensure there
> is no unintentional increase in authorization scope.

It's not clear to me what the equal to or less than language exactly means,
but it's certainly very suggestive, perhaps in a way that isn't good. To
me, it makes it sound like a purpose is a part of a scope, or that there's
a relationship where certain scopes authorize certain purposes. It seems to
me that while the decision to issue a transaction token should take the
scope of the subject token into account, it potentially would also care
[...]

Issue #202, opened 2025-08-26
Consider allowing "expires_in" parameter


PieterKas:
WGLC Feedback from @bc-pi 

https://datatracker.ietf.org/doc/html/draft-ietf-oauth-transaction-tokens-06#section-7.4-3
MUST NOT on expires_in and scope here seems unnecessary and limiting, no? With some reconciling of scope and purp, as alluded to previously, scope could be useful to know without digging into the Txn-Token. And expires_in is RECOMMENDED RFC8693 

Also see:

https://datatracker.ietf.org/doc/html/draft-ietf-oauth-transaction-tokens-06#section-9.1-3 
As the comment above, I disagree with this. Discussing scope and purp in "Txn-Token Lifetime" is also kinda odd.

PieterKas:
Text from RFC https://www.rfc-editor.org/rfc/rfc8693

expires_in
RECOMMENDED. The validity lifetime, in seconds, of the token issued by the authorization server. Oftentimes, the client will not have the inclination or capability to inspect the content of the token, and this parameter provides a consistent and token-type-agnostic indication of how long the token can be expected to be valid. For example, the value 1800 denotes that the token will expire in thirty minutes from the time the response was generated.

From the text in the draft, I think the intent here was that "expires_in" is not needed since the transaction token already contains expiry info.

I feel omitting expires_in was meant to simplify the implementation, but perhaps the better thing to do here is to be silent on "expires_in" so we don't end up overriding a RECOMMENDED in the original spec. 

Agteed on discussing scope and purpose in this section is out of place.

@gffletch and @tulshi - thoughts?

PR #249, opened 2025-09-15
Removal of `purp` claim

gffletch:
Based on conversations and issue comments in issues #178 and #194, this PR renames the `purp` claim in the Transaction Token to `scope`. Clarifying text has been added to make it clear that the value of the `scope` claim in the Transaction Token does NOT need to be the same value requested or present in a provided access_token. 

Please review for clarity of the added text.

I also removed the IANA registry for 'purp'.


draft-ietf-oauth-transaction-tokens.md
```diff
@@ -349,17 +352,18 @@ JWT claims as well as defines new claims. These claims are described below:
 `sub`:
 : REQUIRED A unique identifier for the subject within the context of the `aud` Trust Domain. Unlike OpenID Connect, the `sub` claim is NOT associated with the `iss` claim.
 
-`purp`:
-: REQUIRED A String defining the purpose or intent of this transaction.
+`scope`:
+`scope`:
+: REQUIRED A JSON string as defined in {{RFC8693}}. Note that the value of this claim is determined by the Transaction Token Service and is not required to match the requested scope nor the scope in any supplied external token.
```
PieterKas, line 357:
```suggestion
: REQUIRED Thee scope claim is defined in {{RFC8693}}. Note that the value of this claim is determined by the Transaction Token Service and is not required to match the requested scope nor the scope in any supplied external token.

File: draft-ietf-oauth-transaction-tokens.md [hunk 2]
```diff
@@ -349,17 +352,18 @@ JWT claims as well as defines new claims. These claims are described below:
 `sub`:
 : REQUIRED A unique identifier for the subject within the context of the `aud` Trust Domain. Unlike OpenID Connect, the `sub` claim is NOT associated with the `iss` claim.
 
-`purp`:
-: REQUIRED A String defining the purpose or intent of this transaction.
+`scope`:
```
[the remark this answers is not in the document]
bc-pi: I took me a min to realize that "\`scope\`:" appears twice on consecutive lines (355 and 356 at the time of writing), after which this suggested deletion makes perfect sense

draft-ietf-oauth-transaction-tokens.md
bc-pi: you should even be able to ref the specific section like this:
```suggestion
: REQUIRED The scope claim is defined in {{Section 4.2 of RFC8693}}. Note that the value of this claim is determined by the Transaction Token Service and is not required to match the requested scope nor the scope in any supplied external token.
```

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -349,17 +352,18 @@ JWT claims as well as defines new claims. These claims are described below:
...
-The `purp` claim captures the exact purpose of this particular transaction. This is often much narrower than a scope value issued to an external client. This is due to the fact that in most cases, the authorization model within the Trust Domain is quite different than the authorization model used with clients external to the Trust Domain. To that end, it is intentional to separate the concept of scope (often fairly coarse-grained) used with external clients from the purpose of the transaction used within the Trust Domain. How a given deployment represents the authorization model within the Trust Domain is out of scope for this specification.
+### Scope claim {#scope-claim}
+The `scope` claim captures, as narrowly as possible, the purpose of this particular transaction. The values used for this claim are defined by the Transaction Token Service as representative of the authorization model defined by the Trust Domain. The value may contain a syntax and structure different from, and represent an intent  narrower, than a scope value issued to an external client. How a given deployment represents the authorization model within the Trust Domain is out of scope for this specification.
```
bc-pi, line 366: the syntax and structure shouldn't be different - it's still a  JSON string containing a space-separated list - but the semantics/meaning and actual value might be different. Also tying to not use the word scope with a different meaning in this paragraph. semantics, right? 

```suggestion
The `scope` claim captures, as narrowly as possible, the purpose of this particular transaction. The values used for this claim are defined by the Transaction Token Service as representative of the authorization model defined by the Trust Domain. The value may be literately and semantically different from, and represent an intent narrower, than a scope value issued to an external client. How a given deployment represents the authorization model within the Trust Domain is at its discretion and not prescribed by this specification.
```

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -616,7 +619,7 @@ A Txn-Token is not resistant to replay attacks. A long-lived Txn-Token therefore
 
 Because Txn-Tokens are short-lived, the Txn-Token response from the Txn-Token service does not contain the `refresh_token` field. A Txn-Token cannot be issued by presenting a `refresh_token`.
 
-The `expires_in` and `scope` fields of the OAuth 2.0 Token Exchange specification {{RFC8693}} are also not used in Txn-Token responses. The `expires_in` is not required since the issued token has an `exp` field, which indicates the token lifetime. The `scope` field is omitted from the response in favor of the `purp` claim in the Txn-Token.
+The `expires_in` field of the OAuth 2.0 Token Exchange specification {{RFC8693}} is not used in Txn-Token responses since the issued token has an `exp` field, which indicates the token lifetime.
```
bc-pi, line 622: this whole line should just be removed but that can be deferred to #202

bc-pi (approved):

Issue #195, opened 2025-08-26
Relace StringOrURI with String


PieterKas:
WGLC feedback from @bc-pi 

https://datatracker.ietf.org/doc/html/draft-ietf-oauth-transaction-tokens-06#section-5.2.2 and a few places 
I'm not sure the StringOrURI data type has proven to be real useful. Just say it's a string. Or a URI, if that's needed. But the conditional must be a URI if it has a colon thing isn't something to propagate IMHO. StringOrURI also isn't properly defined or referenced, which wouldn't be an issue if you stop using it.

PR #250, opened 2025-09-16
Replace StringOrURI with string

PieterKas:
See #195 - opting to replace with strings, rather than URIs.


gffletch (approved):

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -740,6 +740,7 @@ The authors would like to thank the contributors and the OAuth working group mem
 # Document History
 {: numbered="false"}
 [[ To be removed from final specification ]]
+* Replace StringOrURI with string [Relace StringOrURI with String](https://github.com/oauth-wg/oauth-transaction-tokens/issues/195)
```
bc-pi, line 743: need (i think) another line or this renders funny but also shouldn't this go under "Since Draft 06"?

Screenshot 2025-09-17 at 2 58 46 PM

tulshi (approved):

tulshi (approved):

Issue #213, opened 2025-08-27
Compare WIMSE arch doc terminology with Transaction token spec


PieterKas:
Feedback from Joe Salowey

Section 4.- Most of the terms in this section overlap with terms we define in [WIMSE arch spec](https://datatracker.ietf.org/doc/draft-ietf-wimse-arch/) as well (Workload, trust domain, External Endpoint, Call Chain, etc).  I don't think there is a lot of divergence, but it would be good to make sure there isn't and maybe make them more common where it makes sense.

tulshi:
@jsalowey would you be able to take a quick look and point out any divergence?

PR #298, opened 2026-01-12
Revise definitions and expand authentication examples

PieterKas:
Updated definitions and descriptions for authorization context, workload, trust domain, and external endpoint. Expanded examples of public-key based authentication methods. see issue #213


jsalowey (commented):
Thank you, this looks good.

PR #319, opened 2026-04-07
Added doc history

tulshi:


gffletch (approved):

gffletch (approved):

draft-ietf-oauth-transaction-tokens.md
bc-pi, line 691: what about this part?

Image

tulshi:
Hi Brian,
Thanks for bringing this to our attention. Please review this PR that
addresses it: https://github.com/oauth-wg/oauth-transaction-tokens/pull/320

Two spurious/orphan lines were addressed in the revision between 07 and 08
and one was addressed in the revision between 06 and 07. I've updated the
history section to reflect it correctly.

Atul

Issue #321, opened 2026-04-08
Security considerations: subject_token validation


dteleguin:
In the case of `subject_token` being an external access token, we could recommend preforming a revocation check, in addition to the mandatory signature and expiration checks. From the discussion with @PieterKas :

> In terms of revoked tokens, it is up to the transaction token service if it checks for that, or how it obtains that information (there is an ever increasing number of ways to find out about revoked tokens, including SSF/CAEP, token status list and token introspection).

Might be worth adding the above to the Security Considerations section.

PR #336, opened 2026-04-16
Add TTS Issuance Policies section

PieterKas:
Added TTS Issuance Policies section detailing issuance decisions and policies for Txn-Tokens. See issue #321


draft-ietf-oauth-transaction-tokens.md
```diff
@@ -586,6 +586,9 @@ A deployment may include multiple instances of a TTS to improve resiliency, redu
...
 
+## TTS Issuance Policies {#tts-issuance-policies}
+TTS token issuance decisions, including whether to issue a Txn-Token and the claims it contains, are determined by deployment-specific issuance policies and business logic. Not every workload that can make a valid request is expected to receive a Txn-Tokens, as doing so may unnecessarily expand the security boundary. For example, a frontend service handling a user-initiated request may be authorized to obtain a Txn-Token carrying specific request context (e.g., a transaction amount), while another service, though authenticated, may not be authorized to request a Txn-Token with that same context. Such policies may vary based on system architecture, workload identity, transaction context, or risk, and may result in different Txn-Token contents or denial of issuance. The definition and management of these policies are deployment-specific and out of scope for this specification.
```
tulshi: The way this is worded "...receive a Txn-Tokens" feels a bit confusing to me, because almost every workload will "receive" a Txn-Token as a part of request processing. I think your intent is to state that "deployments should be highly selective in which workloads can successfully request a Txn-Token." or something like that, right? Can we just say that?

PieterKas: Thanks Atul - I updated the text, but kept the example as I think it will be helpful.

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -589,6 +589,9 @@ A deployment may include multiple instances of a TTS to improve resiliency, redu
...
 
+## TTS Issuance Policies {#tts-issuance-policies}
+A TTS should be selective in which workloads can successfully request a Txn-Token. A TTS does not need to issue a Txn-Token to every workload that requests one. TTS token issuance decisions, including whether to issue a Txn-Token and the claims it contains, are typically subject to deployment-specific issuance policies and business logic. For example, a frontend service handling a user-initiated request might be permitted to obtain a Txn-Token carrying request context such as a transaction amount, while a downstream service in the same Call Chain might not. Such policies may vary based on system architecture, workload identity, transaction context, or risk, and may result in different Txn-Token contents or denial of issuance. The definition and management of these policies are deployment-specific and out of scope for this specification.
```
tulshi: The example could say "API gateway" instead of "frontend service" as a stronger example, because gateways don't iterate rapidly unlike services, which do.

tulshi (approved):

PieterKas:
@tulshi - had to resolve conflicts (doc history) so need another approval....

draft-ietf-oauth-transaction-tokens.md
PieterKas: Agreed that is more illustrative. Update made.

tulshi (approved):

PR #339, opened 2026-04-17
Fix txn token formatting

### LINKED ISSUE #334: Txn-Token vs Txn-Token
in the text and htmlized versions the Txn-Token header name is hard to distinguish from the name name of the thing. Some quotations or other formatting or something might be good in this case. 

https://datatracker.ietf.org/doc/html/draft-ietf-oauth-transaction-tokens-08#name-txn-token-http-header
Image

The html version is a bit better but still
https://www.ietf.org/archive/id/draft-ietf-oauth-transaction-tokens-08.html#name-txn-token-http-header
Image

tulshi:
Closes #334


PieterKas (commented):
@tulshi Should we be using a different header name than Txn-Token? the double quotes may not be enough to avoid confusion.

bc-pi:
> @tulshi Should we be using a different header name than Txn-Token? the double quotes may not be enough to avoid confusion.

I would not take things that far. Just a little visual differentiator is good enough.

PieterKas (approved):

Issue #325, opened 2026-04-08
"Downscoping" transaction tokens to particular target workloads


dteleguin:
In Transaction Tokens, the `aud` claim represents the entire trust domain, which is opposite to OAuth, where it reflects individual resource servers. There is no claim defined that would represent particular workloads as the receivers for the transaction token.

In the absence of the `aud` equivalent having workload-level granularity, how do we implement the "strict audience check" concept from OAuth?

Example: a workload obtains a TraT with audience set to the entire trust domain. Under some circumstances, it might decide to "downscope" the TraT (via replacement request), limiting its "audience" to only selected workloads inside the trust domain, and thus enforcing only a specific request propagation path.

(Credits: Isabella Dix @ Keycloak Dev Day 2026)

ashayraut:
My understanding is that Txn tokens are meant for all internal workloads in trust domain. Once you start minting tokens for specific workload, then its not scalable, because firstly in large service graph, no one knows how the graph changes. The requesting workload may not know if there are 5 internal workloads or 100 that will receive this token. Adding new internal workload will now require Txn-token service or some intermediate workload in call chain (that replaces Txn token) to know the call graph and workloads in that call graph. I'd say, this adds to complication and not a MUST.

On side note, https://www.ietf.org/archive/id/draft-oauth-transactiontokens-bcp-01.html has some best practices outlined.

dteleguin:
Definitely not a MUST, but might be useful for smaller deployments with an established service graph and higher security requirements.

tulshi:
After the discussion in IETF 124, we decided to not work on replacement semantics in this particular draft (see issue #272) . This seems within the scope of replacement, so I'm inclined to not address this in this particular draft.

gffletch:
@dteleguin is it possible to manage the security requirements via the finely-grained scope values? The transaction token `aud` is for the entire trust domain but it's usefulness is only for those services that support the finely-grained purpose of the transaction. 

It is possible to encrypt data in the `tctx` for specific services if there is a concern about leaking of data.

PieterKas:
Based on discussion between myself, @gffletch and @tulshi, transaction tokens are not opinionated about "aud" to restrict se or presentation to specific workloads (doing so would make them into a type of access token as opposed to a mechanism to preserve transaction context).

tulshi:
Closing this, based on the last comment by @PieterKas. Please reopen if this still needs attention.

dteleguin:
[...]

PR #345, opened 2026-04-29
Clarify transaction tokens vs OAuth 2.0 access tokens

PieterKas:
Clarify the distinction between transaction tokens and OAuth 2.0 access tokens, emphasizing their different purposes and security implications. See issue #325

### DIFF
--- a/draft-ietf-oauth-transaction-tokens.md
+++ b/draft-ietf-oauth-transaction-tokens.md
@@ -97,7 +97,7 @@ informative:
 Transaction Tokens (Txn-Tokens) are designed to maintain and propagate user identity, workload identity and authorization context throughout the Call Chain within a trusted domain during the processing of external requests (e.g. such as API calls) or requests initiated internally within the trust domain. Txn-Tokens ensure that this context is preserved throughout the Call Chain thereby enhancing security and consistency in complex, multi-service architectures.
 --- middle
 
-# Introduction
+# Introduction {#introduction}
 
 Modern computing architectures often use multiple independently running components called workloads. In many cases, external invocations through interfaces such as APIs result in a number of internal workloads being invoked in order to process the external invocation. These workloads often run in virtually or physically isolated networks. These networks and the workloads running within their perimeter may be compromised by attackers through software supply chain, privileged user compromise or other attacks. Workloads compromised through external attacks, malicious insiders or software errors can cause any or all of the following unauthorized actions:
 
@@ -597,6 +597,13 @@ The TTS may need to rotate signing keys. When doing so, it MAY adopt the key rot
 
 ## Transaction Tokens Are Not Authentication Credentials
 A workload MUST NOT use a transaction token to authenticate itself to another workload, service or the TTS. Transaction tokens represents information relevant to authorization decisions and are not workload identity credentials. Authentication between the workload and the TTS is described in {{Mutual-Authentication}}.
+
+## Transaction Tokens Are Not OAuth 2.0 Access Tokens
+A workload MUST NOT use a transaction token and an OAuth 2.0 Access Token. An OAuth 2.0 Access Token represents a grant of authority to a client, a transaction token represents the context of a specific transaction. The separation of the transaction token from the OAuth 2.0 Access Token is a deliberate design choice as described in {{introduction}}.
+
+Transaction tokens support the principle of least privilege since they are narrowly scoped to a single transaction and have short lifetimes. This makes them less susceptible to broad replay attacks than the longer-lived access tokens. By maintaining a distinct token type, the protocol ensures that context propagation while minimizing the risk of lateral access if a transaction token is intercepted within the trust domain. This separation also reduces the need for including transaction context in the access token, reducing token size.
+
+Using a dedicated HTTP header from the `Authorization: Bearer` header for transaction tokens allows services to distinguish between an access token, which carries authorization delegation information, and the transaction token, which carries transaction context.
 
 ### Txn-Token as a `subject_token`
 A TTS MUST exercise caution when receiving a Txn-token as a `subject_token`. Any Txn-Token issued in response to such a request is effectively a replacement Txn-Token. Replacing Txn-Tokens potentially negates the primary purpose of having Txn-Tokens. When issuing replacement Txn-Tokens, a TTS:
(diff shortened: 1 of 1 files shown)


ashayraut (approved):

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -589,6 +589,13 @@ The TTS may need to rotate signing keys. When doing so, it MAY adopt the key rot
 ## Transaction Tokens Are Not Authentication Credentials
 A workload MUST NOT use a transaction token to authenticate itself to another workload, service or the TTS. Transaction tokens represents information relevant to authorization decisions and are not workload identity credentials. Authentication between the workload and the TTS is described in {{Mutual-Authentication}}.
 
+## Transaction Tokens Are Not OAuth 2.0 Access Tokens
+A workload MUST NOT use a transaction token and an OAuth 2.0 Access Token. An OAuth 2.0 Access Token represents a grant of authority to a client, a transaction token represents the context of a specific transaction. The separation of the transaction token from the OAuth 2.0 Access Token is a deliberate design choice as described in {{introduction}}.
```
gffletch: I think you meant... "as an" instead of "and an".

gffletch (commented):
Minor typo

draft-ietf-oauth-transaction-tokens.md
```suggestion
A workload MUST NOT use a transaction token as an OAuth 2.0 Access Token. An OAuth 2.0 Access Token represents a grant of authority to a client, a transaction token represents the context of a specific transaction. The separation of the transaction token from the OAuth 2.0 Access Token is a deliberate design choice as described in {{introduction}}.
```

PieterKas: Yup -thanks for catching that. Fixed.

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -589,6 +589,13 @@ The TTS may need to rotate signing keys. When doing so, it MAY adopt the key rot
 ## Transaction Tokens Are Not Authentication Credentials
 A workload MUST NOT use a transaction token to authenticate itself to another workload, service or the TTS. Transaction tokens represents information relevant to authorization decisions and are not workload identity credentials. Authentication between the workload and the TTS is described in {{Mutual-Authentication}}.
 
+## Transaction Tokens Are Not OAuth 2.0 Access Tokens
+A workload MUST NOT use a transaction token as an OAuth 2.0 Access Token. An OAuth 2.0 Access Token represents a grant of authority to a client, a transaction token represents the context of a specific transaction. The separation of the transaction token from the OAuth 2.0 Access Token is a deliberate design choice as described in {{introduction}}.
```
tulshi, line 593: I'm actually rethinking this. I think it makes sense to have OAuth JWT format TraTs that include context (i.e. all the fields that we define in this spec). We need to figure out if we can send such TraTs through the Authorization header or keep it in the Txn-Token header, but adding a sentence to explicitly disallow that sounds a bit extreme to me.

ashayraut: From systems maintenance, observability, separation of concerns and combined with the fact that you may not need Txn token at all on some request chains/paths, it made sense to me always to keep it separate.

PieterKas: What use case do you have in mind @tulshi ?

An OAuth token can contain transaction token claims (a given for the cross-domain use cases), but an OAuth access token is not a transaction token. One is about authorization context (transaction token), the other about broad user access delegation (OAuth).

gffletch: I think in some cases the Authorization header might contain a token that authorizes the client to make the request an not actually be an access token. More of an authentication token but sent in the Authorization header. Like the token returned from a client credentials flow.

I agree that we want transaction tokens in their own header but I think we should be silent on whether an additional token is needed.

tulshi: Should we drop this particular change?

PieterKas: No, I think it is important to be clear that a transaction token is not an OAuth 2.0 Access Token (the question has come up enough times that it is worth being explicit, just as we are about transaction tokens not being used for authentication).

George makes a different point and that is that a Authorisation header might contain a transaction token and that it might be used for authorization (but not as an OAuth 2.0 Access token). 

This text does not say that a transaction token cannot be in the authorisation header as well, just that it should be in its own header and that it is not an OAuth 2.0 JWT Access Token.

ashayraut: Does https://www.ietf.org/archive/id/draft-oauth-transactiontokens-bcp-01.html sound like right place to add clarity around such things?

tulshi: To answer your question @PieterKas , I am now thinking we can include TraTs-like context in access tokens for ensuring AI call chains within an organization do not exceed the bounds of the original task

PieterKas: I think that is fine to include Trats claims in OAuth Access tokens (may even be needed for cross domain use cases when claims have to be transcribed). If it is fine grained authorisation your after, perhaps RAR is the better bet (or perhaps you use RAR and embed Trats defined claims in them),. However, using an Trat claim in an Oauth token does not make a Trat and OAuth Access token. An Oauth Access Token is issued by an AS and conveys delegation semantics. It is used to access the "front door" of the service. Transaction tokens were designed as something else. By merging them, I think you debase the reason for having transaction. tokens in the first place.

tulshi: It isn't an OAuth token in that it hasn't been issued using an OAuth flow, but should we rule out using a TraT as an access token?

PieterKas: We should rule out using a Transaction Token as a OAuth 2.0 Access Token to avoid confusion and dilute the value of both OAuth 2.0 Access Tokens and Transaction Tokens. 

If you have a Transaction Token, and you want an OAuth 2.0 Access Token, use on of the OAuth Authorisation Grant flows (preferably Token Exchange). The draft is clear that a Transaction Token carries transaction specific authorization context to be used as part of an authorization decision. This does not make it an OAuth 2.0 Access Token. It was not issued under the same policies as other OAuth 2.0 Access tokens. Different use cases, threat models and deployment choices apply.

For finer grained OAuth Access tokens, use RAR (perhaps with Transaction token claims). For carrying transaction context, tied to a trasnaction with downscoped authorization details within a domain use a transaction token.

tulshi: I'm not convinced we need this language to avoid confusion, but I'm not too passionate about this, so let's just go ahead with it.

tulshi (approved):

bc-pi:
I support adoption

PR #346, opened 2026-05-13
Change Txn-Token Request parameters to RECOMMENDED

gffletch:
Updated the Txn-Token Request parameters from OPTIONAL to RECOMMENDED to enhance clarity and guidance. Also changed the introduction sentence which previously also included the term RECOMMENDED.

NOTE: Brian also made the recommendation to move from RECOMMENDED to something less like a SHOULD as we don't define concretely what goes in these objects. Additionally, I didn't change body claims as I believe the OPTIONAL make sense for a JWT body claim but less so for an API request parameter. Happy to update accordingly.


PieterKas (approved):
This looks good. I think keeping it to RECOMMENDED as opposed to a SHOULD is fine.

PieterKas (approved):

PieterKas (approved):

tulshi (approved):

PR #354, opened 2026-06-17
moved key determination to security considerations

### LINKED ISSUE #329: iss issue
https://datatracker.ietf.org/doc/html/draft-ietf-oauth-transaction-tokens-08#section-10.2-2.2.1 has "The iss claim MUST be used in cases where the signing keys are not predetermined." but what does the iss claim do in cases where the signing keys are not predetermined? I am very aware that some other specs/profiles/whatever use issuer as part of a key discovery routine. But this isn't any of those specs/profiles/whatever so this **MUST** isn't meaningful.

tulshi:
Closes #329

### DIFF
--- a/draft-ietf-oauth-transaction-tokens.md
+++ b/draft-ietf-oauth-transaction-tokens.md
@@ -296,7 +296,7 @@ The transaction token body follows the JWT format and includes existing
 JWT claims as well as defines new claims. These claims are described below:
 
 `iss`:
-: OPTIONAL The `iss` claim as defined in {{RFC7519}} is not required as Txn-Tokens are bound to a single Trust Domain as defined by the `aud` claim and often the signing keys are known. The `iss` claim MUST be used in cases where the signing keys are not predetermined.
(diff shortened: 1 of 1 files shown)


draft-ietf-oauth-transaction-tokens.md
```diff
@@ -609,6 +609,9 @@ A TTS MUST exercise caution when receiving a Txn-token as a `subject_token`. Any
 * SHOULD limit the number of times a Txn-Token is replaced if it allows extending the lifetime beyond that of the input Txn-Token to reduce replay risks.
 * MUST append the workload identifier of the workload requesting the replacement to the `req_wl` claim using the character `,` as the separator between individual workload identifiers.
 
+## Determining signing keys
+The `iss` claim may be used to identify the signer of the Txn-Token in cases where the signing keys are not predetermined.
```
gffletch, line 613: Do you think we need to define how the 'iss' claim can be used to discover the keys or maybe give an example? Basically something about leveraging .well-known to find the jwks url and accompanying keys??

PieterKas (approved):
Looks good to me

PR #360, opened 2026-06-25
Clarified req_wl value

### LINKED ISSUE #331: ... req_wl ...
https://datatracker.ietf.org/doc/html/draft-ietf-oauth-transaction-tokens-08#section-10.2-2.20.1 

> req_wl:
> 
> REQUIRED. A string value that identifies the workload that requested the Txn-Token. The value SHOULD contain a single workload identifier. In some circumstances, the value MAY contain multiple workload identifiers with each identifier separated by a comma (,).

What are some circumstances? [14.11.1.](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-transaction-tokens-08#section-14.11.1) has two bullets (that should probably be together) about req_wl that are maybe these "some circumstances" but that's a lot left to the reader. Also not great justification for a under-defied structured format for use inside a different data format that has ways to structure data.

What if some of workload identifiers contain commas? Workload identifiers aren't defined in here, are they?

Do we really want to define a sometimes comma separated values list as a the value of a claim?

If the [Requester Context](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-transaction-tokens-08#name-requester-context) claim can contain "information about the computational entity that requested the Txn-Token", then how/why is that different than "the workload that requested the Txn-Token"? 

Could the ["act" (Actor) Claim](https://www.rfc-editor.org/rfc/rfc8693.html#name-act-actor-claim) be of any relevance here? I am painfully aware that representing delegation is a bit of a mess. But does introducing another construct help the situation? I've not yet read @mcguinness's https://notes.karlmcguinness.com/notes/standardize-act-across-assertion-grants-and-jwt-access-tokens/ and don't know if I agree with proposal(s) but highly suspect it describes some of that "bit of a mess". Is a new sometimes-csv valued claim at the top-level going to improve the situation or enworseify it?

tulshi:


draft-ietf-oauth-transaction-tokens.md
```diff
@@ -329,7 +329,7 @@ JWT claims as well as defines new claims. These claims are described below:
 : RECOMMENDED A JSON object that describes the environmental context of the requested transaction. For additional context please refer to {{transaction-context}}
 
 `req_wl`:
-: REQUIRED. A string value that identifies the workload that requested the Txn-Token. In the general case, the `req_wl` value is an array of case-sensitive strings. In the common special case when there is one requesting workload, the `req_wl` value MAY be a single case-sensitive string.
+: REQUIRED. A string or array value that identifies the workload that requested the Txn-Token. In the common special case when there is one requesting workload, the `req_wl` value MAY be a single case-sensitive string. In the general case, when a Txn-Token is replaced by another Txn-Token, the `req_wl` value is an array of case-sensitive strings, identifying subsequent workloads that have requested the Txn-Token. See Section {{replacement}} for more information.
```
```suggestion
: REQUIRED. A string or array value that identifies the workload that requested the Txn-Token. In the common special case when there is one requesting workload, the `req_wl` value MAY be a single case-sensitive string. In the general case, when a new Txn-Token is requested using an existing Txn-Token as a `subject_token`, the `req_wl` value is an array of case-sensitive strings, identifying subsequent workloads that have requested the Txn-Token. See Section {{replacement}} for more information.
```

PieterKas (changes requested):
One suggested change so we don't start talking about replacement, but rather about the case where a Txn-Token is used as a subject token when requesting a new Txn-token

draft-ietf-oauth-transaction-tokens.md
gffletch: Do we need to change this text at all?

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -329,7 +329,7 @@ JWT claims as well as defines new claims. These claims are described below:
 : RECOMMENDED A JSON object that describes the environmental context of the requested transaction. For additional context please refer to {{transaction-context}}
 
 `req_wl`:
-: REQUIRED. A string value that identifies the workload that requested the Txn-Token. In the general case, the `req_wl` value is an array of case-sensitive strings. In the common special case when there is one requesting workload, the `req_wl` value MAY be a single case-sensitive string.
+: REQUIRED. A string or array value that identifies the workload that requested the Txn-Token. In the common special case when there is one requesting workload, the `req_wl` value MAY be a single case-sensitive string. In the general case, when a new Txn-Token is requested using an existing Txn-Token as a `subject_token`, the `req_wl` value is an array of case-sensitive strings, identifying subsequent workloads that have requested the Txn-Token. See Section {{replacement}} for more information.
```
gffletch:
```suggestion
: REQUIRED. An array value that identifies the workload(s) that requested the Txn-Token. In the common case when there is one requesting workload, the `req_wl` value MAY be a single case-sensitive string. In the general case, when a new Txn-Token is requested using an existing Txn-Token as a `subject_token`, the `req_wl` value is an array of case-sensitive strings, identifying subsequent workloads that have requested the Txn-Token. See Section {{replacement}} for more information.
```

PieterKas: Thanks George.

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -329,7 +329,7 @@ JWT claims as well as defines new claims. These claims are described below:
 : RECOMMENDED A JSON object that describes the environmental context of the requested transaction. For additional context please refer to {{transaction-context}}
 
 `req_wl`:
-: REQUIRED. A string value that identifies the workload that requested the Txn-Token. In the general case, the `req_wl` value is an array of case-sensitive strings. In the common special case when there is one requesting workload, the `req_wl` value MAY be a single case-sensitive string.
+: REQUIRED. An string value that identifies the workload that requested the Txn-Token. If a subsequent transaction token is created from a call to the TTS where the subject token is itself a transaction token, the TTS should maintain the call chain of requesting workloads in the subsequently issued transaction token. The mechanisms for managing this delegation chain is out of scope for this specification.
```
```suggestion
: REQUIRED. A string value that identifies the workload that requested the Txn-Token. If a subsequent transaction token is created from a call to the TTS where the subject token is itself a transaction token, the TTS should maintain the call chain of requesting workloads in the subsequently issued transaction token. The mechanisms for managing this delegation chain is out of scope for this specification.
```

PieterKas: I wonder if this can be separated from whether a transaction token is used as subject_token. There are other tokens that may have chained identity information as well (as described in Karl's drafts). How about the following:


```suggestion
: REQUIRED. A string value that identifies the workload requesting the Txn-Token. When the subject token contains information about the call chain, including identifiers for other participants in that chain, the TTS should include this information in the issued Txn-Token. The mechanism for representing and managing the delegation chain is outside the scope of this specification.
```

PieterKas: Given wher this is landing and the source of the concerns, why not just simplify to this:

```suggestion
: REQUIRED. A string value that identifies the workload that requested the Txn-Token. 
```

We can keep the text in the "transaction token as subject token" section, but we break the link here. Feels cleaner and also addresses Brian's other concerns about encouraging the use of replacement transaction tokens.

gffletch: I'm good with that. I think the explanation should go in the transaction token as subject token section.

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -329,7 +329,7 @@ JWT claims as well as defines new claims. These claims are described below:
 : RECOMMENDED A JSON object that describes the environmental context of the requested transaction. For additional context please refer to {{transaction-context}}
 
 `req_wl`:
-: REQUIRED. A string value that identifies the workload that requested the Txn-Token. In the general case, the `req_wl` value is an array of case-sensitive strings. In the common special case when there is one requesting workload, the `req_wl` value MAY be a single case-sensitive string.
+: REQUIRED. A string value that identifies the workload that requested the Txn-Token. If a subsequent transaction token is created from a call to the TTS where the subject token is itself a transaction token, the TTS should maintain the call chain of requesting workloads in the subsequently issued transaction token. The mechanisms for managing this delegation chain is out of scope for this specification.
```
```suggestion
: REQUIRED. A string value that identifies the workload requesting the Txn-Token. When the subject token contains information about the Call Chain, including identifiers for other participants in that Call Chain, the information is preserved in the issued Txn-Token. The mechanism for representing and managing the delegation chain is outside the scope of this specification.
```

PieterKas (changes requested):

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -609,7 +609,8 @@ A TTS MUST exercise caution when receiving a Txn-token as a `subject_token`. Any
 * MUST NOT issue a new Txn-Token when the Txn-Token being replaced has expired.
 * MAY issue a replacement Txn-Token with a lifetime exceeding the lifetime of the input Txn-Token, subject to the policy of the TTS.
 * SHOULD limit the number of times a Txn-Token is replaced if it allows extending the lifetime beyond that of the input Txn-Token to reduce replay risks.
-* MUST append the workload identifier of the workload requesting the replacement to the `req_wl` claim using the character `,` as the separator between individual workload identifiers.
+* When the subsequent transaction token is created, the TTS should maintain the call chain of requesting workloads in the subsequently issued transaction token. The mechanisms for managing this delegation chain is out of scope for this specification.
```
```suggestion
* The TTS should include the Call Chain of requesting workloads in the subsequently issued transaction token. The representation of this Call Chain chain is out of scope for this specification.
```

gffletch (approved):

PieterKas (approved):

PieterKas:
Made one more change to have consistency in terminology and remove ambiguity about removing workload identities from the transaction token.

draft-ietf-oauth-transaction-tokens.md
```diff
@@ -607,11 +607,10 @@ A TTS MUST exercise caution when receiving a Txn-token as a `subject_token`. Any
 * MAY enable additional asserted values.
 * MUST NOT enable modification to asserted values that expand the scope of permitted actions.
 * MUST NOT modify `txn`, `sub`, and `aud` values of the Txn-Token in the request.
-* MUST NOT remove any of the existing requesting workload identifiers from the `req_wl` claim.
 * MUST NOT issue a new Txn-Token when the Txn-Token being replaced has expired.
 * MAY issue a replacement Txn-Token with a lifetime exceeding the lifetime of the input Txn-Token, subject to the policy of the TTS.
 * SHOULD limit the number of times a Txn-Token is replaced if it allows extending the lifetime beyond that of the input Txn-Token to reduce replay risks.
-* MUST append the workload identifier of the workload requesting the replacement to the `req_wl` claim using the character `,` as the separator between individual workload identifiers.
+* MUST maintain the Call Chain of workloads that requested the Txn-Token being replaced, and MUST NOT remove or obscure any of those workload identities in the subsequently issued Txn-Token. The mechanism for maintaining this Call Chain is out of scope for this specification.
```
PieterKas: I would propose dropping "obscure" and just say that is should be maintained (I can imagine some anonymisation may be applied etc, so lets not preclude that and also not introduce a debate about what "obscure" means.

```suggestion
* MUST maintain the Call Chain of workloads that requested the Txn-Token being replaced in the subsequently issued Txn-Token. The mechanism for maintaining this Call Chain is out of scope for this specification.
```

PieterKas (approved):

Issue #349, opened 2026-05-14
More rctx/tctx fun


dteleguin:
From [12.3. Txn-Token Request Processing](https://www.ietf.org/archive/id/draft-ietf-oauth-transaction-tokens-08.html#section-12.3):
> If a `request_context` parameter is present in the Txn-Token Request, the data SHOULD be added to the `rctx` object of the Txn-Token.

Does the "added" here mean "added as is" (copied)? Or is TTS allowed to perform some sort of mapping? The examples in 10.2.4 / 12.1 suggest the former. Would be nice to clarify this.

> If a `request_details` parameter is present in the Txn-Token Request, then the TTS SHOULD propagate the data from the `request_details` object into the claims in the `tctx` object as authorized by the TTS authorization policy for the requesting client.

The wording is different here ("propagated" vs. "added"), so I'd say some sort of mapping is assumed. To further intricate things, there is no `request_details` in the TraT request example in 12.1, making this a pure guesswork. Would be nice to clarify this as well, and also add `request_details` to the example.

Finally, the examples for `rctx` / `tctx` in 10.2.2 / 10.2.3 are inconsistent:

```json
{
    "rctx": {
      "req_ip": "69.151.72.123", // env context of external call
      "authn": "urn:ietf:rfc:6749" // env context of external call
    }
}
```

```json
"tctx": {
  "action": "BUY", // parameter of external call
  "ticker": "MSFT", // parameter of external call
  "quantity": "100", // parameter of external call
  "customer_type": { // computed value not present in external call
    "geo": "US",
    "level": "VIP"
  }
}
```

`rctx` is enclosed in curly braces, whilst `tctx` is not. And ultimately, `quantity` should be a JSON number and not a string (`100` instead of `"100"`).

gffletch:
> If a request_context parameter is present in the Txn-Token Request, the data SHOULD be added to the rctx object of the Txn-Token.

So the intent is that the TTS has full authority to NOT transfer all claims from the incoming request to the `rctx/tctx` claim. Is the following wording more clear?

If a `request_context` parameter is present in the Txn-Token Request, the TTS SHOULD evaluate the content of the `request_context` parameter and propagate appropriate claims to the `rctx` claim of the issued Txn-Token. The TTS is authoritative for the set of claims made available in the `rctx` claim.

This keeps the wording more consistent with that for `request_details`.

dteleguin:
Thanks George, this is definitely more clear.

[...]

PR #361, opened 2026-07-01
Refine tctx and rctx claim handling in Txn-Token

PieterKas:
See issue #349


draft-ietf-oauth-transaction-tokens.md
```diff
@@ -488,8 +489,8 @@ When the TTS receives a Txn-Token Request it:
...
 * The TTS MUST evaluate the value specified in the `scope` parameter of the request to determine the `scope` claim of the issued Txn-Token.
-* If a `request_context` parameter is present in the Txn-Token Request, the data SHOULD be added to the `rctx` object of the Txn-Token.
-* If a `request_details` parameter is present in the Txn-Token Request, then the TTS SHOULD propagate the data from the `request_details` object into the claims in the `tctx` object as authorized by the TTS authorization policy for the requesting client.
+* If a `request_context` parameter is present in the Txn-Token Request, the TTS SHOULD evaluate the content of the `request_context` parameter and may include claims in the `rctx` claim of the issued Txn-Token based on the `request_context` parameter. The claims included in the `rctx` claim of the issued Txn-Token may be an exact match of those on the `request_context`, or may be derived from the `request_context`. The TTS is authoritative for the set of claims made available in the `rctx` claim and MAY include additional claims unrelated to the `request_context`.
```
tulshi: "exact match or derived from request_context" may not be sufficient. The TTS might decide to add values of its own in the rctx". Perhaps we can say "exact match, derived from the request_context or independently asserted by the TTS"

PieterKas: I think that is what this sentence tries to do:

"The TTS is authoritative for the set of claims made available in the `rctx` claim and MAY include additional claims unrelated to the `request_context`."

But I like your suggestion as it is more concose and all in the same sentence.

```suggestion
* If a `request_context` parameter is present in the Txn-Token Request, the TTS SHOULD evaluate the content of the `request_context` parameter and may include claims in the `rctx` claim of the issued Txn-Token based on the `request_context` parameter. The TTS is authoritative for the set of claims made available in the `rctx` claim. The claims included in the `rctx` claim of the issued Txn-Token may be an exact match of those on the `request_context`, may be derived from the `request_context`, or independently asserted by the TTS.
```

tulshi (approved):

PR #366, opened 2026-07-29
Refine description of Txn-Token JWT body claims

PieterKas:
Clarified the description of Txn-Token body claims and their handling.


gffletch:
I approve. This is good clarification text. For some reason I can't approve it (at least not from the iOS GitHub app).

gffletch (approved):

